PR #44, opened 2024-05-27
fix linux build

ricewind012:


src/core/backend/co_spawn.cpp
```diff
@@ -27,6 +27,8 @@ plugin_manager::plugin_manager()
     console.log_item("status", "done appending init tabs!");
     console.log_item("status", "initializing python...");
 
+    Py_Initialize();
```
ShadowMonster99, line 30: Py_Initialize() shouldn't be called. it is already called on L:63 with config. This is most likely why you are having to manually specify paths.

src/core/backend/co_spawn.cpp
```diff
@@ -35,19 +37,35 @@ plugin_manager::plugin_manager()
...
+    ssize_t list_size = PyList_Size(sys_path);
+
+    for (size_t i = 0; i < list_size; i++) {
+        PyObject* item = PyList_GetItem(sys_path, i);
+        ssize_t item_size = PyList_Size(item);
+        wchar_t* str = PyUnicode_AsWideCharString(item, &item_size);
+
+        PyWideStringList_Append(&config.module_search_paths, str);
+    }
+    Py_XDECREF(sys_path);
```
ShadowMonster99, line 56: this snippet requires python to be initialized, before its actually initialized. This is most likely why you needed Py_Initialize() beforehand (which is also invalid)

src/core/backend/co_spawn.cpp
```diff
@@ -35,19 +37,35 @@ plugin_manager::plugin_manager()
...
+
+#ifdef _WIN32
     PyWideStringList_Append(&config.module_search_paths, std::wstring(pythonPath.begin(), pythonPath.end()).c_str());
     PyWideStringList_Append(&config.module_search_paths, std::wstring(pythonLibs.begin(), pythonLibs.end()).c_str());
+#endif
 
     status = Py_InitializeFromConfig(&config);
+    if (PyStatus_Exception(status)) {
+        console.err("couldn't read config");
```
ShadowMonster99, line 65: error message should be more descriptive. "failed to initialize from config"

ShadowMonster99 (changes requested):
changes required before merge

ShadowMonster99 (approved):

ShadowMonster99 (approved):

PR #64, opened 2024-07-17
Drastically reduce cloned repo size

AHOHNMYC:
Cloned only current repo state without history. 0.5 MB instead of 28 MB.


ShadowMonster99 (approved):
Thanks, this is a great optimization I completely overlooked!

PR #109, opened 2024-09-17
Check if jq is installed

shvedes:
I have repeatedly encountered the problem after a clean installation of Arch, when a script tries to run a not installed jq. For some reason, the script did not check whether it was installed or not. On Arch, for example, this package is not included in the standard package groups.


shdwmtr:
Thanks!

PR #128, opened 2024-10-10
Removed the need for pnpm

FenrirBots:
Updated gitignore to ignore generated files/folders.
Updated package.json to build using npm instead of pnpm.
Updated README.md to show npm instead of pnpm


shdwmtr:
Why? PNPM is being used because its better than NPM. Maybe there is something I'm missing?

FenrirBots:
Well it just seems odd to use pnpm when most people have nodejs installed already

FenrirBots:
got busy but i looked it up and i see why pnpm is here now

PR #138, opened 2024-10-16
Fix couldn't not resolve (pipx.config)

ZeroNiki:
import pipx.config as config could not resolved

pipx/config.py pipx/main.py


shdwmtr:
Show me the error you receive in the logs.

PR #171, opened 2024-11-19
Close console after install

SpaceEnergy:
When the installation is complete, the window is automatically closed.

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -73,14 +73,6 @@ function Start-Steam {
     }
 
     Start-Process -FilePath $steamExe -ArgumentList "-verbose"
-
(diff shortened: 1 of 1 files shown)


shdwmtr (changes requested):
This is counter intuitive though, if it closes the script when steam is started, the logs will never display millennium starting up. This means there could be an error the end-user would never see.

To fix this, you can add a check `$message` on L:441 to see if Millennium has logged `CORE SteamUI successfully loaded!` (with a regex match), and if it has that means Millennium has successfully loaded, and closing the power shell window is safe.

shdwmtr (approved):
Looks good, I'll merge it once I get to test it.

PR #178, opened 2024-12-10
Update install.sh

Bread-Ch4n:
Change line [#4](https://github.com/shdwmtr/millennium/blob/8d8e6e3dd499b639583d00cf816c7d4f2fe90a42/scripts/install.sh#L4) to say Linux and not Windows


shdwmtr:
Oops, thanks!

PR #184, opened 2024-12-14
Fix settings icons not being centered and plugin settings button not having propper styling

tddebart:
Just a simple fix for the settings icons not being centered
  
Before:
before themes
After:
after themes

Before:
before plugins
After:
after plugins


shdwmtr:
Thanks! I see you've come across the `renderPluginSettings` function (it seems), which is generally unfinished. You can implement portions of it if you need such a function, and I'll try to get the rest done (but don't feel obligated to).

Cheers!

Issue #227, opened 2025-02-07
[Feature] Follow FHS standard


Sk7Str1p3:
### Before Requesting

- [x] I found no existing issue matching my feature request
- [x] This request is related to Millennium, and not a request for a specific theme/addon

### Describe the feature you'd like!

I tried packaging Millennium for NixOS but found what is not following FHS standard and cant really understand which files can be safely put in r/o store and which are not. if you cant or dont want to, i would do that, but i must know file structure. thx

shdwmtr:
Millennium on linux is still somewhat of a work in progress, so its FHS isn't final. If there is a standard somewhere I can look into it to move Millenniums assets around to where they are more useful.

Sk7Str1p3:
i suggest this layout:
```
 /
├──  bin
│   ├──  .millennium-unwrapped (executable)
│   └──  millennium (start.sh)
├──  lib
│   ├──  libMillennium.so
│   └──  libpython-3.11.8.so
│       └── (should not be bundled with millennium imo. better set python as dependency)
└──  home
    └──  user
        ├──  .config
        │   └──  Millennium
        │       ├──  plugins
        │       ├──  millennium.ini
        │       └──  themes.json
        └──  .local
            └──  share
                └──  Millennium
                    └──  logs
                        ├──  core_log.txt  
                        ├──  pacman.log 
                        └──  pipx_log.txt

```

i cant move files from 'assets', 'shims' and 'cache' folders yet. What are they supposed for? may you explain what is contained there @shdwmtr ?

shdwmtr:
I'd have to change the source code to accommodate such changes. They are core runtime dependencies for Millennium, it won't work without them.

Sk7Str1p3:
I don't tell you to kick em out. I just have no idea where could we place it. I will try to restructure release and send you .tar

Sk7Str1p3:
alr so this is wanted layout, needed for ability to package Millenium as native pkg for any distro like NixOS, Arch Linux, OpenSUSE, Debian, etc.:

```
 .
├──  bin
│   ├──  millennise
│   └──  millennium
├── 󱂵 home
│   └──  user
│       ├──  .config
│       │   └──  Millennium
│       │       ├──  millennium.ini
│       │       ├──  plugin.json
│       │       └──  themes.json
│       └──  .local
│           └──  share
│               └──  Millennium
│                   └──  logs
│                       ├──  core_log.txt
│                       ├──  pacman.log
[...]

PR #230, opened 2025-02-10
Check Python package on Linux systems

mourogurt:
Some basic changes to better find Python libs on Linux system. Also remove /usr as output dir as the user in most cases doesn't have write permission in it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,47 @@ endif()
 project(Millennium LANGUAGES CXX)
 add_subdirectory(cli)
 
+# Check Python version in 32 bit section
+# Just include Python headers for Windows and Apple platforms
+if(WIN32)
+  include_directories(${CMAKE_SOURCE_DIR}/vendor/python/include)
+elseif(UNIX)
+  if(APPLE)
+    include_directories("$ENV{HOME}/.pyenv/versions/3.11.8/include/python3.11")
+  else()
+    # Try to find required version of Python
+    # Python guarantee to have API and ABI compatible within a same major and minor versions
+    # Harden version range to to find only 3.11 python
+    find_package(Python 3.11 EXACT REQUIRED COMPONENTS Development)
+
+    if (PYTHON_FOUND)
+      # Run simple test to check if python is working with current flags
+      try_compile(PYTHON_TEST_RESULT
+        "${CMAKE_BINARY_DIR}"
+        SOURCES "${CMAKE_CURRENT_LIST_DIR}/tests/FindPython_test.cpp"
+        LINK_LIBRARIES Python::Module)
+        if(PYTHON_TEST_RESULT)
+          message(STATUS "Found suitable Python version ${Python_VERSION}")
+          set(LIBPYTHON_RUNTIME_PATH ${Python_LIBRARIES})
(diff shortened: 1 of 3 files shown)


Sk7Str1p3:
@shdwmtr as I promise pull request with build and FHS fixes (special thanks to @mourogurt)

shdwmtr:
This doesn't solve any issues. This puts the version management of python on the system, which could lead to undefined behavior. 

> @shdwmtr #227 : If it offers 32 bit python 3.11 (Steam is 32bit) it should work, but I can't recommend something like that, let alone let other users do the same.

* Not all / most package managers don't offer 32bit python 3.11.8, which is a required standard for Millennium. 
* The package isn't built specifically for Millennium (i.e may not contain all the same site-packages etc)

For instance, if you build Millennium with Python 3.12.3 and use a plugin made for 3.11.8, its possible something in the plugin was deprecated, and crashes if you try it on 3.12.3.

Regardless, no checks are made if this is being configured on NixOS, so these changes affect all distrobutions.

mourogurt:
Will it be all good if cmake will check python for specific version (with Python_VERSION variable) and check for bitness (probably with try_compile)? If anything fails, then it will fallback to the ~/Documents/LibPython location.

mourogurt:
There is no such directory as `/usr/lib` in NixOS, hence this is why all builds must be done in special environments in that distro (it creates symlinks and modifies rpath to make sure every library will search for the right version of dependency).

mourogurt:
Or maybe it's better to set PYTHONHOME environment variable to find libs and include pathes, and use it in the code to find `libpython.so`?

Sk7Str1p3:
> There is no such directory as `/usr/lib` in NixOS, hence this is why all builds must be done in special environments in that distro (it creates symlinks and modifies rpath to make sure every library will search for the right version of dependency).

Not required. We can simply move libs to /lib in installPhase

shdwmtr:
> Will it be all good if cmake will check python for specific version (specify in find_package and with Python_VERSION variable) and check for bitness (probably with try_compile)? If anything fails, then it will fallback to the ~/Documents/LibPython location. IF all check passes, then cmake will create define a macro that represents path to a libpython.so , else it will just be `/usr/.../libpython.so.3.11.8`

Sounds like a feasible idea, I can't think of anything wrong with that off the top of my head.

mourogurt:
Done and done.

mourogurt:
> Does NixOS have a 32 bit build of Python?

Yes, you can use pkgsi686Linux to get 32 bit pacakges inside the shell like this:
```nix
{
  pkgs ? import <nixpkgs> {},
  pkgsi686Linux   ? import <nixpkgs> { system = "i686-linux"; }
}:
....
buildInputs = with pkgsi686Linux; [ python311 ];
....

````
> This needs to be updated to use LIBPYTHON_RUNTIME_PATH

> As for the test file, the rest of the project uses the .cc naming standard so it should probably also follow that

Will fix tommorow

> Additionally, If my memory isn't mistaking me, this isn't building as 32bit:

try_compile inherits all the CMAKE_* flags, threrefore I just move section up before -m32 been dropped. But will add CMAKE_FLAGS for extra safety.

PR #232, opened 2025-02-11
Updated Russian Translation

dotFelixan:
There may be mistakes because I don't know the context.


shdwmtr:
I'll leave this open until I release the new update, then it will be easier to understand the context just in case you need to update something.

dotFelixan:
> I'll leave this open until I release the new update, then it will be easier to understand the context just in case you need to update something.

I think it's worth adding for testing, anyway the update is still a "Actions" Build.

I check all new builds and compare them with what I translated

PR #234, opened 2025-02-12
Install Millenium library in default installation path

mourogurt:
In one configuration (when we use bundled python package for Linux system) CMake doesn't rely on CMAKE_RUNTIME_OUTPUT_DIRECTORY to install libraries. Therefore, add installation step to copy that library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -214,3 +214,5 @@ elseif(UNIX)
     endif()
   endif()
 endif()
+
+install(TARGETS Millennium DESTINATION lib)
--- a/cli/CMakeLists.txt
+++ b/cli/CMakeLists.txt
@@ -25,3 +25,5 @@ add_compile_definitions(MILLENNIUM_VERSION="${MILLENNIUM_VERSION}")
 
 set_target_properties(cli PROPERTIES OUTPUT_NAME "millennium")
 set_target_properties(cli PROPERTIES PREFIX "")
+
+install(TARGETS cli DESTINATION bin)


mourogurt:
During installation phase of CMake package, Nix runs command `cmake --build . --target install --prefix "some path"` and fails, because there are no install targets in Millennium. This small change fixes it.

shdwmtr (changes requested):
EDIT: I realized you're using `lib` not `/lib` so its entirely possible it means something other than what I thought. 

Changes made are NixOS specific, and wont work for other distributions.


I recommend

`/cmake/distro.cmake`

```cmake
set(OS_RELEASE_PATH "/etc/os-release")

if(NOT EXISTS "${OS_RELEASE_PATH}")
    message(WARNING "${OS_RELEASE_PATH} not found, unable to detect Linux distribution")
    return()
endif()

file(READ "${OS_RELEASE_PATH}" OS_RELEASE_CONTENTS)
string(REGEX MATCH "ID=([a-zA-Z0-9_-]+)" MATCHED_ID "${OS_RELEASE_CONTENTS}")

if(MATCHED_ID)
    set(DISTRO_ID "${CMAKE_MATCH_1}")
    message(STATUS "Detected Linux Distribution: ${DISTRO_ID}")

    if(DISTRO_ID MATCHES "ubuntu|debian|pop|mint|elementary|raspbian|kali|slackware|alpine|void|bodhi|deepin|mageia|zorin")
        set(IS_DEBIAN_BASED TRUE)
        message(STATUS "Detected Debian-based distribution")
        set(DISTRO_BIN_PATH "/usr/bin")
        set(DISTRO_LIB_PATH "/usr/lib")

    elseif(DISTRO_ID MATCHES "arch|manjaro|endeavouros|artix")
        set(IS_ARCH_BASED TRUE)
        message(STATUS "Detected Arch-based distribution")
        set(DISTRO_BIN_PATH "/usr/bin")
        set(DISTRO_LIB_PATH "/usr/lib")

    elseif(DISTRO_ID MATCHES "fedora|rhel|centos|rocky|almalinux")
        set(IS_FEDORA_BASED TRUE)
        message(STATUS "Detected Fedora/RHEL-based distribution")
        set(DISTRO_BIN_PATH "/usr/bin")
        set(DISTRO_LIB_PATH "/usr/lib")

    elseif(DISTRO_ID MATCHES "opensuse|suse")
        set(IS_OPENSUSE_BASED TRUE)
        message(STATUS "Detected openSUSE-based distribution")
        set(DISTRO_BIN_PATH "/usr/bin")
        set(DISTRO_LIB_PATH "/usr/lib")

    elseif(DISTRO_ID MATCHES "gentoo")
        set(IS_GENTOO_BASED TRUE)
        message(STATUS "Detected Gentoo-based distribution")
        set(DISTRO_BIN_PATH "/usr/bin")
        set(DISTRO_LIB_PATH "/usr/lib")

    elseif(DISTRO_ID MATCHES "nixos")
        set(IS_NIXOS TRUE)
        message(STATUS "Detected NixOS")
        set(DISTRO_BIN_PATH "/bin")
        set(DISTRO_LIB_PATH "/lib")

    else()
        set(IS_UNKNOWN_LINUX TRUE)
        message(STATUS "Unknown Linux distribution: ${DISTRO_ID}")
        set(DISTRO_BIN_PATH "/usr/bin")
        set(DISTRO_LIB_PATH "/usr/lib")
    endif()

    message(STATUS "Binary path:  ${DISTRO_BIN_PATH}")
    message(STATUS "Library path: ${DISTRO_LIB_PATH}")
endif()
``` 

`CMakeLists.txt`

```cmake
include(cmake/distro.cmake)
```


And using `${DISTRO_BIN_PATH}`, `${DISTRO_LIB_PATH}` , `${DISTRO_ID}` wherever needed.

Sk7Str1p3:
@shdwmtr just noticed i have to pass additional flag to found Python.h which is included in repo. error not appears with other headers

mourogurt:
@Sk7Str1p3 It's separate issue and maybe related to build recipe and not to source code.

mourogurt:
> EDIT: I realized you're using lib not /lib so its entirely possible it means something other than what I thought.

So what does DESTINATION do in install command: it will create corresponding folder to the CMAKE_INSTALL_PREFIX (which is --prefix "some path" in cmake command) and install target (library/executable in corresponding places).

If I write:
```bash
$ cmake --install . --prefix ~/package 
-- Install configuration: ""
-- Installing: /home/desktop/package/lib/libMillennium.so
-- Installing: /home/desktop/package/bin/millennium
```
It will create ~/package/bin and ~/package/lib directories (if not exist), and copy corresponding targets to them.
Likewise, if I write: `cmake --install . --prefix /usr`
CMake will install millenium to the user folder.

mourogurt:
About NixOS: it's create folder in /nix-storage/<some-hash><package> for each package, where it will download or install (if no binary find in repository, then it will run cmake build + install) it.

mourogurt:
See: https://cliutils.gitlab.io/modern-cmake/chapters/install/installing.html

shdwmtr:
> Maybe it's convinient to remove `CMAKE_RUNTIME_OUTPUT_DIRECTORY` and `LIBRARY_OUTPUT_DIRECTORY` entirely and add chapter to the README.md file about installing Millenium from sources?

It wouldn’t be convenient on my end, I use that to develop Millennium. With the way I have it right now, All I have to do is build and restart Steam.

mourogurt:
So, will it be fine to accept PR as is then?

PR #257, opened 2025-03-09
Add linux build inside docker container

LazyHatGuy:


shdwmtr:
Don't know much about docker, but what does this provide that isn't already there? The linux version is already built from the CI.

LazyHatGuy:
It allows you to build the latest version locally

PR #262, opened 2025-03-09
Add linux build inside docker container

LazyHatGuy:
The rebase went a bit wrong, so recreating the pull request

The advantage of this, being able to build the Linux version locally while keeping the build process containerized


cli/CMakeLists.txt
```diff
@@ -18,12 +19,19 @@ add_compile_definitions(
   "FMT_HEADER_ONLY"
 )
 
+FetchContent_Declare(fmt
+  GIT_REPOSITORY https://github.com/fmtlib/fmt.git
+  GIT_TAG master
+)
+FetchContent_MakeAvailable(fmt)
+
```
shdwmtr, line 27: `FetchContent` is unneeded. It was just a bug I fixed a couple commits ago with

`include_directories(../vendor/fmt/include)`

shdwmtr (changes requested):

docker-build-linux.sh
```diff
@@ -0,0 +1,130 @@
+#!/bin/bash
+
+# Adapted from install.beta.sh by LazyHatGuy
```
shdwmtr, line 3: Missing some information here. Since this is maintained by you, put an explicit GitHub URL / email denoting you maintain it. You'll also need some sort of license that you find fits, it just can't clash with the parent projects MIT Licensing. 

ex: https://github.com/shdwmtr/millennium/blob/main/src/main.cc
But In your case have explicit contact information since you aren't a public maintainer.

shdwmtr (changes requested):

docker-build-linux.sh
LazyHatGuy: I like your standards for FOSS and as someone who works in FOSS approvals I should have thought of this

shdwmtr:
Any progress on this?

LazyHatGuy:
I'm going by the SPDX standard https://spdx.org/licenses/MIT.html

PR #283, opened 2025-03-25
fix: Use XDG base directory environment variables

### LINKED ISSUE #282: [Bug] XDG base directory variables not being used
### Before Reporting

- [X] I found no existing issues matching my bug
- [X] My issue is not caused by a theme
- [X] My issue is not caused by a plugin

### Describe the Bug

(the errors are irrelevant cuz I moved my ~/.local/share folder to ~/usr/share)
image

### Expected Behavior

Millennium config/data files being in $XDG_CONFIG_HOME/$XDG_DATA_HOME respectively.

### Steps To Reproduce

1. Set $XDG_CONFIG_HOME, $XDG_DATA_HOME, etc. to something else.
2. See Millennium not using it.

### Operating System

Linux

### Anything else?

_No response_

ricewind012:
fixes #282

### DIFF
--- a/pkgbuild/PKGBUILD
+++ b/pkgbuild/PKGBUILD
@@ -41,7 +41,8 @@ build() {
 
 package() {
     cd "$srcdir/millennium"
-    destinationBase="$pkgdir$HOME/.local/share/millennium/lib/assets"
+    local dataPath="${XDG_DATA_HOME:=$HOME/.local/shrae}"
+    local destinationBase="$pkgdir$dataPath/millennium/lib/assets"
 
     declare -A paths=(
         ["./assets/.millennium/Dist/index.js"]="$destinationBase/.millennium/Dist/index.js"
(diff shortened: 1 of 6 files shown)


pkgbuild/PKGBUILD
```diff
@@ -41,7 +41,8 @@ build() {
 
 package() {
     cd "$srcdir/millennium"
-    destinationBase="$pkgdir$HOME/.local/share/millennium/lib/assets"
+    local dataPath="${XDG_DATA_HOME:=$HOME/.local/shrae}"
```
shdwmtr, line 44: Typo

pkgbuild/millennium.install
```diff
@@ -1,23 +1,26 @@
 #!/bin/bash
 
+: "${XDG_DATA_HOME:=$HOME/.local/share}"
```
shdwmtr, line 3: not familiar with this syntax, is this setting XDG_DATA_HOME? Because if not directly below you reference XDG_DATA_HOME without a fallback to $HOME

src/sys/env.cc
```diff
@@ -109,19 +115,22 @@ const void SetupEnvironmentVariables()
...
         { "LIBPYTHON_RUNTIME_PATH",  LIBPYTHON_RUNTIME_PATH },
 
-        { "MILLENNIUM__STEAM_EXE_PATH", fmt::format("{}/.steam/steam/ubuntu12_32/steam",     std::getenv("HOME")) },
-        { "MILLENNIUM__PLUGINS_PATH",   fmt::format("{}/.local/share/millennium/plugins",    std::getenv("HOME")) },
-        { "MILLENNIUM__CONFIG_PATH",    fmt::format("{}/.config/millennium",                 std::getenv("HOME")) },
-        { "MILLENNIUM__LOGS_PATH",      fmt::format("{}/.local/share/millennium/logs",       std::getenv("HOME")) },
-        { "MILLENNIUM__DATA_LIB",       fmt::format("{}/.local/share/millennium/lib",        std::getenv("HOME")) },
-        { "MILLENNIUM__SHIMS_PATH",     fmt::format("{}/.local/share/millennium/lib/shims",  std::getenv("HOME")) },
-        { "MILLENNIUM__ASSETS_PATH",    fmt::format("{}/.local/share/millennium/lib/assets", std::getenv("HOME")) },
+        { "MILLENNIUM__STEAM_EXE_PATH", fmt::format("{}/.steam/steam/ubuntu12_32/steam",     homeDir) },
```
shdwmtr, line 127: If I remember correctly, Steam doesn't abide by `XDG_DATA_HOME` so if its set, and differs from `HOME` it would point to the wrong steam directory.

shdwmtr (changes requested):

pkgbuild/millennium.install
ricewind012: it sets XDG_DATA_HOME to $HOME/.local/share if unset or empty

src/sys/env.cc
ricewind012: that uses homeDir, not dataDir

shdwmtr: Oh whoops, no idea how I misread that so badly.

shdwmtr (approved):

PR #286, opened 2025-03-28
feat: Add steam://open/millennium URL to open settings

ricewind012:


shdwmtr:
I know you're still adding something to this (I think) but I'm going to merge this because your plugin is broken on the new version of Millennium, and I'm releasing it right now. I want users to have a way to get to Millennium settings.

PR #295, opened 2025-03-30
feat: steam://millennium URL support

ricewind012:


shdwmtr:
Does this fix the issue of not being able to click "View Profile" in the friends list?

ricewind012:
yeah that was cuz of the "open" part in steam://open/millennium

PR #318, opened 2025-04-12
refactor: Move all Millennium styles to a separate file

ricewind012:
- move all styles to src/styles.tsx (styles.css preferably but idk if i can use my own rollup plugins)
  - there, move everything about colors/text/spacing into variables, which are also separated
- try to integrate with the rest of steam settings more
  - remove bottomSeparator Field prop usage where appropriate
  - the \<pre> element in the logs tab now uses "DialogInput DialogTextInputBase" in order to distinguish it from the rest
- add a "data-default-color" boolean attribute to the accent color field in the themes tab
  - remove the "Reset" button if default color is in use
- new ButtonsSection component to put your buttons in there if there are 2+ of them
- apply the "active" (same style as hover) class name for the millennium context menu to highlight it instead

and other shit i forgot to write about


shdwmtr:
I'll take a look into this soon. `styles.tsx` is probably more preferable anyways because you can use Steams dynamic classes.

PR #322, opened 2025-04-18
Fix millennium hanging up on linux when steam exits

Kaosume:
This fixes https://canary.discord.com/channels/1102739071085846623/1358686818693808128, also fixes -s flag not applying thus saving ~1.2 mb on my machine.


shdwmtr:
I'll have to test this; it's setup like that for a reason. Due to underlying differences with how the kernels handle threads on Windows vs. Linux, you're setup might break thread ownership guidelines that are important for unloading Millennium on windows. 

Either way, thanks for contributing! I'll take a look ASAP.

PR #329, opened 2025-04-26
fix(security): only allow whitelisted paths via file proxy url

BossSloth:
Fixed a security issue allowing file access to the whole system.
Also added some more filetypes for images to the file proxy


shdwmtr:
Some little nitpicks
* `add_proxy_whitelisted_regex_path` && `remove_proxy_whitelisted_regex_path` are probably a bit too a verbose. 
* Make sure to follow Allman style C++ guidelines. 
* `EscapeRegex` should probably be moved elsewhere (maybe it's own header) as it's pretty generic and isn't specific to the parent class. 
* `m_whiteListedRegexPathsPtr` should be stack allocated. Although that only saves like 90 bytes of memory its better practice. 

Other than that, great work :) (assuming it works)
I'll merge this after I merge next into main.

BossSloth:
Fixed the first 3 nitpicks.
`m_whiteListedRegexPathsPtr` needs to be a shared_ptr otherwise it doesn't get updated if called from `_c_py_api.cc` just like the `m_hookListPtr` next to it or there might be something that I am missing.

src/core/web_load.cc
```diff
@@ -121,7 +161,7 @@ MILLENNIUM void WebkitHandler::RetrieveRequestFromDisk(nlohmann::basic_json<> me
     }
 
     uint16_t    responseCode    = bFailedRead ? 404 : 200;
-    std::string responseMessage = bFailedRead ? "millennium" : "millennium couldn't read " + localFilePath.string();
```
shdwmtr, line 124: Lol, no idea how this slipped into prod.

shdwmtr:
Those two are slightly different. Since `m_whiteListedRegexPathsPtr` is instantiated **and assigned/given a memory address** within a header file, when the header is copied over to each source file when compiling, there will actually be two different symbol table entries for `m_whiteListedRegexPathsPtr` after each source file is [mangled](https://en.wikipedia.org/wiki/Name_mangling); making them essentially two completely different variables that don't point to the same memory address. 

To fix this, you can stack allocate `m_whiteListedRegexPathsPtr` and assign it inside the `web_load.cc` source file instead of inline. This way only one instantiation of `m_whiteListedRegexPathsPtr` occurs as source files aren't copied.

Alternatively, you can C style top level declarations `extern`ing it in the header file and linking it in the source file. 

As per why `m_hookListPtr` is a shared pointer and not stack allocated is because the lifespan of `m_hookListPtr` shouldn't be bottle-necked by the webkit handler class as although its defined within itself, it isn't **owned** by itself and needs to stay alive throughout the entire runtime (or as long as the python API has reference to it) to prevent nullptr assignment.

I know this all probably seems like dumb semantics but even a tiny mistake like this left forgotten can be the reason peoples Steam crash. C++ / C were flawed languages they tried to fix over the years, but only made it more confusing lol.

BossSloth:
that is really stupid. Fixed it now by just putting it as an extern in the header file as i could not get it to work inside of the class

shdwmtr:
Isn't C++ fun :)

shdwmtr (approved):

shdwmtr:
Needs documentation before merging. As its a pretty breaking change, it needs to be in the docs.

BossSloth:
Added documentation https://github.com/SteamClientHomebrew/SteamBrew/pull/15

BossSloth:
@shdwmtr this pr seems to have been auto closed when you deleted the `next` branch(and i can't reopen it). 
But is this pr still needed or has it become obsolete because of the fixes that have been done for [#367](https://github.com/SteamClientHomebrew/Millennium/issues/367#issuecomment-2985482455)?  
As currently it seems like the FTP is still accessible from everywhere and still giving access to all files.

ejalxndr:
Those changes were to prevent external browser connections (i.e Chrome, Firefox) to Millennium, not the internal store pages, community, etc. So we still need a fix for that. I just don't know if we should use this as a solution. I don't currently have a better solution on hand, but this just feels like band-aiding the root issue instead of stitching it closed.

BossSloth:
aha, yeah I wasn't fully happy with this fix either but I also couldn't come up with anything better. 
Guess I'll just leave it closed for now.

PR #348, opened 2025-05-28
Update russian.json

seropheem:
Added missing lines.


ejalxndr:
This PR will need an update, many things are now behind!

PR #350, opened 2025-05-28
improved webkit load performance by ~350ms

BossSloth:
Improved load time till `ExecutePluginModule` for the plugin webkit is called.
Before: ~1100ms
After: ~750ms
Load time was measured on an app page by adding a `performance.mark` at the start `ExecutePluginModule` and grabbing the event log time from the performance tab in the devtools averaged over 8-10 loads restarting the client twice in between (because for some reason steam's load times vary between boot ups)

It looks like there is not much more time to save since you are dealing with the main thread being full with parsing html or calling steam functions most functions calls or async responses are just pending because the main thread is just busy.

### DIFF
--- a/shims/src/index.ts
+++ b/shims/src/index.ts
@@ -85,7 +85,7 @@ function formatCssVarKey(key: string): string {
 
 const StartPreloader = async (port: number, shimList?: string[]) => {
 	logger.log(`Successfully bound to ${isClient ? 'client' : 'webkit'} DOM...`);
-	const socket = await CreateWebSocket('ws://localhost:' + port);
+	const socket = await CreateWebSocket('ws://127.0.0.1:' + port);
 
 	/** Setup IPC */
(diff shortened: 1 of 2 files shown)


shdwmtr:
What appears to be taking so much time? Network tab says it took 3ms to download the webkit.js, and 3ms to connect to Millennium's backend.

I must be missing something here.

BossSloth:
It is not the network or the c++ side that is the problem, it is just that the js main thread is way too busy at the start of the page load. Connecting to the ipc takes less than 1 ms if the thread is free. But because other things like parsing the html get priority it needs to wait on them.
You can take a look for it yourself. Load this file into the chrome profiler then you can see that every timing event is happening when there is a small break between parsing html or other big function calls. It seems to be kinda random when it will call some function or resolve them.
[Trace-20250528T191828.json](https://github.com/user-attachments/files/20504050/Trace-20250528T191828.json)

PR #358, opened 2025-06-10
Fix use of chown

RozeThorns:
chown would try to use the user's name as both the group and user which would cause an error
this approach gets the primary group of the user instead of assuming both are the same

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,11 +1,9 @@
 #!/usr/bin/env sh
-
 # This file is part of the Millennium project.
 # This script is used to install Millennium on a Linux machine.
 # https://github.com/shdwmtr/millennium/blob/main/scripts/install.sh
 # Copyright (c) 2025 Millennium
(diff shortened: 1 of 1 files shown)


ejalxndr:
This script is targeted towards modern linux distributions like Ubuntu, Debian, Fedora, Arch, that use [UPG](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/4/html/reference_guide/s1-users-groups-private-groups). In a standard linux system, 

> A UPG has the same name as the user for which it was created and that user is the only member of the UPG. 

In this case, no issues would arise.

NihonFemStyle:
this had caused me an error since i've actually changed the name on my user account which is why it actually caused me an issue, dont know if its still worth pulling but my honest opinion is its best to be ready for anything

ejalxndr:
Millennium no longer writes to home directory so this can now be closed :)

PR #359, opened 2025-06-11
Linux fixes

Kaosume:
This will fix first LD_PRELOAD error message, millennium looking for possibly wrong steam installation (you looked for ~/.local/share/Steam in one part of the code and for ~/.steam in another, i made that it looks only for ~/.steam), install.sh was not checking for git which is needed for core plugin at least.
Also do not change start.sh to your local millennium build: change your /usr/bin/steam or create copy elsewhere.
I think ifdef way more foolproof as it will check if path is defined, also you defined assetPath twice and broke linux build, (and i made release fix before i saw you fixed it and i like my way better :) ).
If you needed those var ping me and i'll undo last commit.

### DIFF
--- a/src/sys/env.cc
+++ b/src/sys/env.cc
@@ -102,7 +102,6 @@ const void SetupEnvironmentVariables()
         { "MILLENNIUM__STEAM_PATH",   SystemIO::GetSteamPath()  .string() }
     };
 
-    #ifdef MILLENNIUM_SDK_DEVELOPMENT_MODE_ASSETS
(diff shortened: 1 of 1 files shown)


ejalxndr:
I just committed your `env.cc` fix right into next. I had changed some stuff previously so I thought I'd just commit it directly to make it easier for you to remove it from your PR. the rest should be good though.

Kaosume:
Should i rebase and remove commits to env.cc or it's all good?

ejalxndr:
Oh oops, sorry I forgot to update you. I fixed it.

PR #363, opened 2025-06-13
Allow millennium to restart steam using different port

Kaosume:
This is not ready:
- It is based of main
- idk if startupParams are collected
- messages are wacky
- choice saved in file, might be needed to be saved in options
- there might be some caveats to using exec on windows
- there's no error handling
 
but the idea is that we say that the port is taken and prompt user about it, then we restart steam with command line arg of the new port and if user checked this will happen w/o prompt: user starts steam -> port taken -> millennium restarts stream with new port.
It'll prob slow down startup but it'll allow seamless port change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -305,7 +305,7 @@ find_package(CURL REQUIRED) # used for web requests.
 target_link_libraries(Millennium CURL::libcurl)
 
 if(WIN32)
-  target_link_libraries(Millennium wsock32 Iphlpapi DbgHelp)
+  target_link_libraries(Millennium wsock32 Iphlpapi DbgHelp comctl32)
 
   if(GITHUB_ACTION_BUILD)
     target_link_libraries(Millennium "D:/a/Millennium/Millennium/Python-3.11.8/PCbuild/win32/python311.lib")
--- a/include/await_pipe.h
+++ b/include/await_pipe.h
@@ -41,6 +41,12 @@
 #include <tlhelp32.h>
 #include <psapi.h>
 #include <commctrl.h>
+#include "cmd.h"
+#include "locals.h"
+const TASKDIALOG_BUTTON portdialogbtns[] = {
+    {301, L"Restart Steam with different port"},
+    {302, L"Exit"},
+};
 #elif __linux__
 #include <fstream>
 #include <sstream>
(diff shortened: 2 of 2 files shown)


ejalxndr:
I personally feel like this is a bad way to address the issue. We used to have a solution like this a year back, but removed it because its not a clean solution, and there is likely too many moving parts to make this consistent. 


I have somewhat of an idea, however I've never made a POC. It might be possible to trampoline version.dll and hook `GetCommandLineW`, to change the port.

ejalxndr:
Make a POC, after reverse engineering Steam I found it wasn't possible to hook GetCommandLineW, so I opted for a different solution. 

I noticed steam load-time links to `version.dll` from System32 (essentially meaning it loads version.dll before main) but they don't absolutely link to its System32 path meaning I can intercept it with a DLL forwarder from the current working directory. 

I reversed version.dll and found it had some functions Steam relied on, so I pipe forwarded them to the original version.dll like
`some_function="C:\\Windows\\System32\\version".some_function`, and then it was possible to make my own DllMain to hook calls before steam ever runs a line of code. 

I was then reversing how `-devtools-port` is serialized in steam, and found it originates from ordinal 9 in `steamclient.dll`. I then found Steams `tier0_s.dll` cross platform handler has a function `Plat_CommandLineParamValue` which `steamclient.dll` calls to parse the `-devtools-port` command line argument. 

{0ECF75A0-7A42-4CFD-8274-4A4355479ED2}

`Plat_CommandLineParamValue` is forwarded to, and dynamically linked to `tier0_s.dll` aswell meaning its possible to trampoline its pipe call without inline patching its IAT within `steamclient.dll`.

So I hooked `LdrRegisterDllNotification` from `ntdll` to intercept the callback on when modules are loaded into the current process, and it also luckily forwards the base address of the module that loads. So I wait for `tier0_s.dll` to load into memory, and then trampoline `Plat_CommandLineParamValue` with minhook. 

I made a typedef for `Plat_CommandLineParamValue` but not entirely sure if its right (pretty sure it is though, but needs review). Under the standard 32bit calling convention `__cdecl` `[EBP + 8]` stores the first parameter and `[EBP + 12]` is the second, and since there isn't a second parameter (just first parameter and -4 return jump address) it seems it doesn't accept a default param. 

{E1D5FA8E-02B9-4C02-B05F-89A66ECFD838}

Anyways, not sure if this will ever make its way into Millennium, but you can check it out 
https://github.com/shdwmtr/devtools-port-hook

PR #375, opened 2025-06-22
fix external python path

Sk7Str1p3:
fixed by @mourogurt


Sk7Str1p3:
WRONG BRANCH IM SORRY

PR #376, opened 2025-06-22
set MILLENNIUM__PYTHON_ENV when external python is used

Sk7Str1p3:
fixed by @mourogurt


shdwmtr (approved):

mourogurt:
I only proposed the idea with `Python_ROOT_DIR` and doesn't test it at all. It turns out to be empty.
We need to add something like this:
```cmake
if(NOT Python_ROOT_DIR)
   cmake_path(GET Python_LIBRARY_DIRS PARENT_PATH Python_ROOT_DIR)
endif()
```

It will get root path from lib path.

After that, millenium successfully launched on my machine:
```
[12:06.445] CORE Checking for updates...
[12:06.916] CORE Millennium is ahead of the latest release.
[12:06.916] BOOT Creating __millennium_plugin_settings_parser__ function in builtins.
[12:06.916] CORE Loading Millennium-Core@v2.26.0
[12:06.916] INFO Successfully loaded 'core'
```

@Sk7Str1p3 Can you create new PR with this fix?

Sk7Str1p3:
Yes, I will

PR #399, opened 2025-07-02
Handle arrays in Python FFI

Functionable:
Seems like `InvokePythonFunction` at one point passed arrays from JS to Python correctly and Non-Steam Playtimes relied on that. This gets the plugin to work again.

### DIFF
--- a/src/core/_c_py_interop.cc
+++ b/src/core/_c_py_interop.cc
@@ -265,6 +265,47 @@ PyObject* resolveDottedAttribute(const std::string& dottedPath)
     return current;
 }
 
+/*
+ * Returns a representation of the given JSON data as a PyObject*
+ *
+ * @param json The JSON object to be converted
+ * @return A PyObject* representing the passed JSON object
+ */
+PyObject* PyObjectFromJson(const nlohmann::json& json)
+{
+    PyObject* pObject = nullptr;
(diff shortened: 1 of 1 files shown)


ejalxndr:
## Issues

* If `PyList_SetItem` fails, you return `nullptr` which skips clean up on the partially constructed list. This leaks memory from `PyList_New` calls. 
* Inconsistency with var name? Unless I'm missing something here (I haven't actually test compiled) `counter` is undefined.
* A `LOG_ERROR` call should probably be used if the array parse fails. 
* Another issue I also just noticed (which is my fault), I accidentally truncate a `long` to an `int` here `pObject = PyLong_FromLong(json.get<int>());`, so we should fix that while we're at it! (`PyLong_FromLongLong(json.get<long long>());` might be a good fix for that)


Thanks for contributing!

Functionable:
I renamed `counter` to `i` just before sending this off initially, and I clearly forgot to compile to check. I'm a bit embarrassed by that.
Fixes for the issues to do with the `PyList_SetItem` call have been rolled into a single commit which should probably be fine.

I also fixed the truncation issue by doing exactly what you suggested.

The code should compile fine this time. I double checked this time...

Thank you very much for your feedback.

PR #402, opened 2025-07-03
fix nix build

Sk7Str1p3:


ejalxndr:
You don't necessarily need to use git patches, you can simply make a compiler derivative like `NIX_OS` similar to how we have `_WIN32`, `__linux__`, etc. It would make your build far more stable as you wouldn't need to update your patches every time something near it changes.

shdwmtr (approved):

Sk7Str1p3:
I never worked with C++, not sure how can I do that. 
e.g in `Rust` I would simply use `env!` macro to access build time env variable. How do I implement same?

PR #405, opened 2025-07-04
Create turkish

CoffeemanHell:


ejalxndr:
Re-target the pull request to `next` branch, not `main`. This way we can get these changed out into beta!

ejalxndr:
your PR is behind next, and is offering to update install.sh to a version it already is. This will mess up git history. 
The easiest solution would probably be to make a new PR on next.

PR #406, opened 2025-07-05
Cmake configuration instead of patches for nix build

Sk7Str1p3:


ejalxndr:
Your fork also seems to be behind `next`.

Issue #436, opened 2025-07-21
[Bug] [NixOS] pnpm.fetchDeps returned hash mismatch


Fivemilacc:
### Before Reporting

- [x] I found no existing issues matching my bug
- [x] My issue is not caused by a theme
- [x] My issue is not caused by a plugin

### Describe the Bug

When using Millennium as a Nix flake input, nix fails to build Millennium (next branch), stating that the pnpm dependencies return a different hash than specified.
```error: hash mismatch in fixed-output derivation '/nix/store/jvbkhk4443l6mgcab118sp9in3i4byrn-millennium-assets-pnpm-deps.drv':
         specified: sha256-nDSltpFQRM9loVuDour4OrRdN22/A7MkZTGAtL0x7rU=
            got:    sha256-M2kvId8f1kV8gHSJWubNAS3u0IFrHWKfeci2zrj5lmQ=
error: 1 dependencies of derivation '/nix/store/88l7zs7zwf7shbsirq2qparzpq1zvknh-millennium-assets-git.drv' failed to build
error: 1 dependencies of derivation '/nix/store/dv3bv921frp3cfsj1y0nwiw15zx7s7ml-millennium-git.drv' failed to build
error: 1 dependencies of derivation '/nix/store/xkjjx69h8c84bsad98pgwlg1fyfii9x6-steam-fhsenv-profile.drv' failed to build
error: 1 dependencies of derivation '/nix/store/ygi52z5dkcfsdw91wp800c8jrk86kvjd-steam-run-fhsenv-profile.drv' failed to build
error: 1 dependencies of derivation '/nix/store/hdh0n89vg0gijmd16h0dihb2ndxkcmww-steam-fhsenv-rootfs.drv' failed to build
error: 1 dependencies of derivation '/nix/store/n6sg1276qpjwzx57bh9ymyxli3b3lrm1-steam-run-fhsenv-rootfs.drv' failed to build
error: 1 dependencies of derivation '/nix/store/8lfsgzmr7nw0jgnqpypnm35mmdq5k8cf-steam-bwrap.drv' failed to build
error: 1 dependencies of derivation '/nix/store/508vlns3z4ishs68kd45whkxiw6xqr29-steam-run-bwrap.drv' failed to build
error: 1 dependencies of derivation '/nix/store/9ihz65mi11rvl974f8y9g337xrgphm6s-steam-run.drv' failed to build
[...]

PR #439, opened 2025-07-22
Revert "fix: update pnpmDeps hash"

Fivemilacc:
Reverts SteamClientHomebrew/Millennium#437
Turns out #436 was just my configuration being weird, so reverting this should fix it for others.


shdwmtr:
Just waiting on @Sk7Str1p3's review

Sk7Str1p3 (approved):
I can't check if this works bc I'm away from pc for a while. But, others says the thing was working before this PR, so yes we should revert it. 

@shdwmtr we should also setup workflow for automatic hash update

PR #447, opened 2025-07-28
small fix italian

MakinDay:


ejalxndr:
Target `main` instead of `next`

PR #467, opened 2025-08-13
fix: plugin updater merge instead of replacing

### LINKED ISSUE #466: [Bug] Plugin updater deletes additional files in plugin directory
### Before Reporting

- [x] I found no existing issues matching my bug
- [x] My issue is not caused by a theme
- [x] My issue is not caused by a plugin

### Describe the Bug

Upon updating a plugin, any additional files that that plugin creates, are deleted.
For example, my non-steam-playtimes, upon updating, Millennium deletes the `sessions.json`.

https://github.com/user-attachments/assets/c98fb547-51e9-42be-9ba9-902c5bebfd55

### Expected Behavior

It should not delete/result in the deletion of additional files created in the plugin directory...

### Steps To Reproduce

See video

### Operating System

Windows

### Anything else?

_No response_

apteryxxyz:
Currently the updater deletes the existing plugin, replacing it with the updated version.
You've said in Discord that it shouldn't outright remove anything, so I've made this merge the directories instead to follow that

Closes #466

https://github.com/user-attachments/assets/5fe8f38a-dee1-4a9b-b0ab-2a6b1f92ffe9

### DIFF
--- a/assets/core/plugins/plugin_updater.py
+++ b/assets/core/plugins/plugin_updater.py
@@ -116,10 +116,12 @@ class PluginUpdater:
             extracted_folder = extracted_dirs[0]
             target_folder = Path(self.plugins_path) / name
 
-            # Replace existing plugin folder
-            if target_folder.exists():
-                shutil.rmtree(target_folder)
-            extracted_folder.rename(target_folder)
+            # Merge the extracted folder with the target folder
+            if not target_folder.exists():
+                target_folder.mkdir(parents=True)
+            shutil.copytree(extracted_folder, target_folder, dirs_exist_ok=True)
+            # Cleanup
+            shutil.rmtree(extracted_folder)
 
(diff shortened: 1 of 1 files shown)


ejalxndr:
Does shutil.copytree overwrite existing files?

k0d13:
> Does shutil.copytree overwrite existing files?

Yes, see the file timestamps in the video updating, expect for the sessions.json file
I opened plugin.json just to be sure, it has "version": "1.3.0" (also checked nested files, they are updated as well)

k0d13:
Here's how I tested, manually moved an old version of non-steam-playtimes, deleting the existing folder. Added my backed up sessions.json, open Steam and ran the updater.

If you wanna test, here is an old version of my plugin that I downloaded from the website before 1.3.0 was merged.
[non-steam-playtimes.zip](https://github.com/user-attachments/files/21758574/non-steam-playtimes.zip)

PR #469, opened 2025-08-16
Update brazilian.json

bathtimethiago:


shdwmtr (changes requested):
The JSON is invalid. Also remove the Brazilian flag from `strWelcomeModalTitle`. I have nothing against the flag or country, but just for consistency with other languages.

bathtimethiago:
It’s ok now?

ejalxndr:
No the JSON is still invalid. Put it in https://jsonlint.com/ you can see where the issue is.

bathtimethiago:
I hate , 😡

PR #470, opened 2025-08-16
Delete assets/src/locales/locales directory

bathtimethiago:
Why does this exist?


ejalxndr:
Don't really know lol, must have merged a pr without checking where it was writing to.

PR #479, opened 2025-08-24
Add Serbian (Cyrillic & Latin) locales

rammba:
Hello @shdwmtr, I'm happy to be able to contribute to Steam related project 🚀
I'm currently working on translating the [official React documentation](https://github.com/reactjs/sr.react.dev) to Serbian, but I also contributed with translations to few other open-source projects.

I'm adding two versions of Serbian language, Cyrillic and Latin. Please let me know if I missed to configure or translate something, or if the file naming is incorrect.


ejalxndr:
Is Serbian even supported by Steam?

rammba:
Ahh I didn't think about that. On Steam, there is only Serbian language as a [Game Support Only](https://partner.steamgames.com/doc/store/localization/languages). Does Millenium languages need to be in the correlation with the Steam official languages?

ejalxndr:
Yes, Millennium uses Steam's in built language manager, and if the language isn't apart of Steam's supported languages it won't work.

rammba:
So language needs to be in the `Full Platform Supported Languages` list on the Steam? If that's the case I would mention that in the README.
P.S. Is it possible to merge Serbian language files without including them to the `Millennium` app? I would still try to update these files based on the english.

ejalxndr:
Yeah it should be mentioned in the README; I've just never had a situation were a language wasn't supported on Steam so I never thought of it. 

> So language needs to be in the `Full Platform Supported Languages` list on the Steam?

Yes.

> P.S. Is it possible to merge Serbian language files without including them to the Millennium app? I would still try to update these files based on the english.

They wouldn't be usable though?

rammba:
I've added changes for supported languages in #482.

> They wouldn't be usable though?

Yes, but it will be ready if Steam adds Serbian officially at any moment. If you think that's not a good idea, I'm also ok with that. Please close the PR in that case.

ejalxndr:
I'll leave it here for the time being.

ejalxndr:
I'm going to close this then. If Steam ever adds support for these locales, I'll open this again.

PR #487, opened 2025-08-30
Ensure git exists in Steam's FHSEnv/Various fixes and additions for the Flake

DrymarchonShaun:
So, originally this was just going to be a pr to add git to steam's `extraPkgs` because its not included by default and if the user doesn't have it installed (or steam is launched from somewhere that doesn't have git included in the `PATH`, like a systemd service), GitPython freaks out because it can't find the `git` executable and Millennium's core plugin fails to load.

As I was adding that, a few other small things jumped out at me, (ie, the pnpmDeps hash for the SDK was out of date since the sdk has been updated).

I also saw the TODO for automating the pnpmDeps hashes, while I'm not sure the best/easiest way to fully automate it, I put together a (probably janky) script that *should* update both hashes just by running `nix run .#update-pnpm-hashes` although `nix-update` does seem rather inconsistent so YMMV.

Each of the various changes has its own commit, if anything needs to be changed or removed LMK.

PR #499, opened 2025-09-11
Removed Reliance on Submodules

Trivaris:
Removed any references to /sdk, /vendor/asio /vendor/crow, etc and replaced them with flake inputs

This makes the usage of this flake as an input of another flake possible and simplifies the process of moving nix files into a subdirectory:

```nix
inputs.millennium = {
  url = "github:SteamClientHomebrew/Millennium"
  inputs.nixpkgs.follows = "nixpkgs";
};
```

This also benefits this repo in the way that the gitmodules' revs are unified in one lock file `flake.lock`
The paths.patch file was moved into the derivations, as it was not compatible with the current nixified inputs
> Please let me know if this causes issues!

Lastly, this change also provides easier access when building from source, as submodules do not need to be cloned, its just
`git clone git@github:SteamClientHomebrew/Millennium`
`nix build .#`
and the shared library file can be found in `./result/lib/millenium/libmillennium_x86.so`

I also changed a few bad practices like using `rec` or `with`


DrymarchonShaun:
The only problem with this that I see is that it adds more to the maintenance burden, as it's another thing that needs to be manually updated when the submodules are updated (along with the `shims` package's `pnpmDeps`)

ejalxndr:
The nix build uses vcpkg right? It's possible that all submodules could just be converted to vcpkg deps. I don't know much about how the nix side of this repo works though

Trivaris:
Sorry for this, I made some changes i was not happy with and will come back with a cleaner, more focused solution that will do nothing more than move to inputs and fix any errors arising with that, as otherwise it would be a duplicate of #487

PR #503, opened 2025-09-21
feat: Add atomic distribution support

mrvictory1:
This PR adds support for Linux systems where /usr is read only by utilizing /usr/local which is writable on Universal Blue based distributions. The PR has been tested on Bazzite. List of changes:

* `cli/CMakeLists.txt`: C++ standard has been upgraded from C++ 11 to C++ 17. This is necessary to use std::filesystem.
* `cli/src/main.cc`: new logic has been added to check if /usr/bin/steam is modifiable. If it is not, an attempt is made to create a copy at `/usr/local/bin`. If it is successful, patching is performed in `/usr/local/bin/steam`. If it fails, millennium exits.
* `scripts/install.sh`: Script checks if /usr/bin is writable, if not it /usr/local/bin is checked. If that also fails, script exits.
* `scripts/posix/start.sh`: .so file is searched in both /usr/local/bin and /usr/bin. Lack of file in either directory is harmless.
* `src/sys/env.cc`: If `/usr/local/bin/millennium` exists, /usr/ is changed to /usr/local (done 3 times)

Todo / Caveats:
* /usr/local is not writable on some distributions ie bootc based CentOS. I did not know this when starting working on this. ~/.local/bin can be used instead.
* check_patch_status() in millennium cli does not check for /usr/local.
* Bazzite uses a custom script `/usr/bin/bazzite-steam` to launch Steam on desktop. This needs to be changed to `/usr/bin/local/steam` or simply `steam` which is not prevented by read only /usr.

### DIFF
--- a/cli/CMakeLists.txt
+++ b/cli/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.10)
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 
 # Set the project name and language
--- a/cli/src/main.cc
+++ b/cli/src/main.cc
@@ -42,6 +42,7 @@
 #include <fcntl.h>
 #include <pwd.h>
 #include <iostream>
+#include <filesystem>
 
 #include "incbin.h"
 #include "chown.h"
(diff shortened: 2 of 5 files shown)


ejalxndr:
We probably shouldn't use filesystem in the CLI. its a huge dependency and likely brings significant overhead onto the stack which would increase the binary size.

Pretty sure `copy_file_range` or `sendfile` would work just as well!

timelord1102:
Cant seem to reproduce an install on Bazzite. Files are placed properly by install.sh, but start.sh does not actually apply millennium, instead opening steam with no millennium options.

mrvictory1:
@timelord1102 exit steam fully then start steam from the terminal with `/usr/local/bin/steam` or just `steam`. See point
> Bazzite uses a custom script /usr/bin/bazzite-steam to launch Steam on desktop. This needs to be changed to /usr/bin/local/steam or simply steam which is not prevented by read only /usr.

PRAlSEGABEN:
I merged this into the latest version, and it still doesn't work. No matter how I launch Steam, it just does not start with Millennium.

ninetailedtori:
Not really ideal behaviour here. The only thing that goes into /usr/local/lib/ is globally compiled manual-compiled projects, which are user-installed, as opposed to system-managed packages. It would be incorrect behaviour to do such. There should be a different solution to this, but if nothing else is possible, then we'd have to consider whether we allow going against xdg freedesktop spec for that directory, as far as I know.

ejalxndr:
This PR needs major changes to work with the new source tree.

ejalxndr:
Closing as  stale

PR #519, opened 2025-10-18
fix: fixed quickCss failing when user is not on the millennium alpha version

BossSloth:
applyQuickCss would fail if the backend function does not exist (as it only exists currently on the alpha version) this would make all of the webkit scripts not load in.
This fixes that.
I have tested this locally


ejalxndr:
The backend function [does exist](https://github.com/SteamClientHomebrew/Millennium/blob/main/assets/core/util/ipc_functions.py#L37C1-L39C70), I just accidentally JSON.parse()'d a string 🤦‍♂️. Thank you for the PR regardless! I've applied your try catch aswell as removing JSON.parse

PR #532, opened 2025-10-26
fix(nix): update pnpm dependency hashes and add auto-update workflow

luvcie:
Fixes the Nix build failures caused by outdated pnpm dependency hashes. 

When pnpm dependencies are updated, the Nix hashes must also be updated to match. Previously this was done manually, leading to frequent breakage for NixOS users. 
The new GitHub Action automates this process.

### DIFF
--- a/assets/pnpm-lock.yaml
+++ b/assets/pnpm-lock.yaml
@@ -3078,3 +3078,4 @@ snapshots:
       immer: 10.1.1
       react: 17.0.2
 
+# test


luvcie:
Closing this PR to test the GitHub Action workflow properly in my fork first. GitHub Actions are new to me and I thought the workflow would trigger in the PR to showcase that it works, but I need to test it separately before submitting. Will reopen once I've confirmed everything works correctly.

Issue #527, opened 2025-10-24
[Bug] NixOS. error: attribute 'steam-millennium' missing / Plus error: hash mismatch in fixed-output derivation


KyrTeF:
### Before Reporting

- [x] I found no existing issues matching my bug
- [x] My issue is not caused by a theme
- [x] My issue is not caused by a plugin

### Describe the Bug

On https://docs.steambrew.app/users/getting-started/installation#linux
it is told to use
`
nixpkgs.overlay = [
  inputs.millennium.overlays.default
];
`

However I was getting error: `error: attribute 'steam-millennium' missing`

so I found a fix to add letter s at end of  nixpkgs.overlay
from there on I could start system rebuild with correct code:
`
nixpkgs.overlays = [
  inputs.millennium.overlays.default
];
`

Here I could start rebuild but run into an error
`
error: hash mismatch in fixed-output derivation '/nix/store/f946lfpkdajzjpmj3sx3j5w3wgnv7kgl-millennium-assets-pnpm-deps.drv':
         specified: sha256-nDSltpFQRM9loVuDour4OrRdN22/A7MkZTGAtL0x7rU=
            got:    sha256-/FR6hFs/JGBAQEp58dw/DUvD1lTo0u5IBFCiveskgBc=
`
such error was mentionet in
#505 however with different hashes
https://github.com/SteamClientHomebrew/Millennium/issues/504#issuecomment-3423947458
with exactly the same hashes.
#436 with different hashes and mentioned issue is inside of assets.nix with faulty hash.

### Expected Behavior

Everything should work.

### Steps To Reproduce

Follow the guide at
https://docs.steambrew.app/users/getting-started/installation#linux

### Operating System

Linux

ghost:
Problem with hash missmatch comes from 2 files.
https://github.com/SteamClientHomebrew/Millennium/blob/00418518da713f292dac23be62d0803cfe9d56c1/nix/assets.nix#L13
https://github.com/SteamClientHomebrew/Millennium/blob/00418518da713f292dac23be62d0803cfe9d56c1/nix/typescript/shims.nix#L15

these two files causing issues becouse they are not autoupdated, therefore for each new build they require new hash. So Fix is obvious: add an autoupdate of these 2 hashes.

ghost:
I propose 2 solution for my Issue. 
1. Update https://docs.steambrew.app/users/getting-started/installation#linux
[...]

PR #534, opened 2025-10-27
fix(nix): update pnpm dependency hashes and add auto-update workflow

luvcie:
This PR fixes hash mismatches that prevented Nix builds from succeeding.

## Changes
Updates millennium-assets pnpm-deps hash in `nix/assets.nix`, updates millennium-sdk pnpm-deps hash in `nix/typescript/shims.nix`, and adds a GitHub Actions workflow to automatically update hashes when pnpm-lock.yaml files change.

## Problem
Users trying to build Millennium with Nix were encountering hash mismatch errors because the pnpm dependency hashes in the Nix files were outdated.

## Solution
This updates both hashes to their correct values and adds a GitHub Actions workflow that will automatically update these hashes in the future whenever the pnpm-lock.yaml files change, so this won't happen again.

## Testing
I tested the workflow in my fork, it successfully detects hash mismatches, extracts the correct hashes from build errors, and commits the updates automatically.

### DIFF
--- a/pkgbuild/PKGBUILD
+++ b/pkgbuild/PKGBUILD
@@ -7,7 +7,7 @@ url="https://github.com/SteamClientHomebrew/Millennium"
 license=('MIT')
 depends=('git' 'steam')
 makedepends=('npm' 'curl' 'zip' 'unzip' 'tar' 'cmake' 'ninja' 'lib32-gcc-libs' 'pnpm')
-depends_x86_64=('python-i686-bin')
+depends_x86_64=('lib32-python311-bin')
 source=("git+$url.git#commit=0cf351871e1c4715e6b4ded3d5df614455fe4af7")
 sha256sums=('SKIP')
 options=(!debug)


DioKyrie-Git:
Closes #527

ejalxndr:
Still references `assets/pnpm-lock.yaml` (old) instead of `src/frontend/pnpm-lock.yaml` (new)

nix/assets.nix
DioKyrie-Git: Should this still be assets everywhere?

nix/typescript/shims.nix
DioKyrie-Git: #TODO seems to be fixed already? By your .github/workflows/update-nix-hashes.yml

luvcie: Yes, that comment is from the previous nix maintainer, I forgot to remove it.

nix/assets.nix
luvcie: You're right, the directory moved from assets to src/frontend in the staging branch, I can update that path.

However I just realized that the staging branch doesn't have Nix support at all yet, while main does. My original intent was just to fix the outdated hashes in main and also put a github actions workflow as a long term fix, but I was asked to target the staging branch instead. 

I think for now this PR would have been better into main as a quick fix.

DioKyrie-Git: The hashes were updated by my pr already, as for autoupdate. Yes this is really needed in main, I thnink you should just leve this PR here for staging branch and make new fork for main branch with only the hash autoupdate workflow. Or whatever works best

DioKyrie-Git:
Why was this sent to staging branch anyway?

luvcie:
> Why was this sent to staging branch anyway?

I reminded @shdwmtr of my PR on discord and they told me "target the staging branch instead of main".
I didn't realize staging was so different.

nix/assets.nix
luvcie: I've changed the PR to target main now instead of staging. 
Since you've already fixed the hashes, this PR now only adds the auto-update github actions workflow. Should be good now. :)

PR #558, opened 2025-12-07
Support for Clang compiler (unwrap variadic arguments), fix code errors thrown by -Wpedantic.

ninetailedtori:
both gcc and clang pass this flag through to linker, so this is for greater compatibility as well.

### DIFF
--- a/src/include/millennium/backend_init.h
+++ b/src/include/millennium/backend_init.h
@@ -83,9 +83,6 @@ class BackendCallbacks : public Singleton<BackendCallbacks>
         PluginTypeSchema(const std::string& name, eBackendLoadEvents ev) : pluginName(name), event(ev)
         {
         }
-        PluginTypeSchema(const PluginTypeSchema& other) : pluginName(other.pluginName), event(other.event)
-        {
-        }
     };
 
     using EventCallback = std::function<void()>;
--- a/src/include/millennium/logger.h
+++ b/src/include/millennium/logger.h
@@ -102,5 +102,16 @@ extern OutputLogger Logger;
 #endif
 
 #ifndef LOG_ERROR
-#define LOG_ERROR(fmt, ...) Logger.ErrorTrace(fmt, __FILE__, __LINE__, PRETTY_FUNCTION, ##__VA_ARGS__)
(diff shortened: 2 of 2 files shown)


ninetailedtori:
Bump to C++20 needed so that Clang can go with `-Wpedantic` due to the issue of `-Wgnu-zero-variadic-macro-arguments`, which is only solvable in C++20 with the addition of `__VA_OPTS__`.

ejalxndr:
Clang support is definitely a good idea, thanks!

However, regarding C++ 20, I'd personally prefer to stay on C++17 as I'd technically have to document the migration as a breaking change (as those who compile from source may need a different/updated version of their compiler) which I prefer to avoid. Is it possible without updating to C++20?

ninetailedtori:
I'd have to use a macro hack that expands based on count of variadics, I think? Lemme see what I can do ;]

ninetailedtori:
Alright, there's some messy ways we could do it - if we attempt to helper function to trick preproc into believing there's always one arg, which there is but, then it'll throw an error in pedantic on variadic-arguments-omitted, which is a valid warn as there'll be a hanging comma. We can instead, define an overload to this macro, which will be my first attempt to see if that works, and that'll do until we bump to c++20, or we have to force gcc, which is messy and sorta ruins the point of this PR xD

I'll attempt the overload, but if that fails then I will rewrite the function to handle variadic unwrapping in c++ native, rather than c preproc, and I'll migrate the code to using that safely instead ( which is technically the best safe solution, just a teeny bit less speed if I don't optimise it right but I'll make sure to optimise it well :] )

ninetailedtori:
But I've cloned the original c++20 migration as well so we have a backup once we do bump to it, so we can revert to speedy preproc (unless I work out a genius native solution, I'll try my best xD)

ninetailedtori:
Alright, I've force pushed a change for this - replaces the entire commit history with a change to VA_ARGS expansion, as well as remove user-defined copy constr for the implicit copy constr which is exactly the same (improves compatibility with strict and pedantic errors, because it throws an error that you've defined one and it's replaced the user-defined copy operator as well).

PR #561, opened 2025-12-08
Git pnpm cmake task. Also add .idea(prevent breaking+override on JetBrains IDEs.)

ninetailedtori:


ejalxndr:
PNPM Changes: 
* I plan to remove pnpm from cmake scripts, or at least have a flag like -DBUILD_ASSETS. 

Git Changes: 
* I plan to completely move the sdk into this repo, so we wouldn't need submodules at all. 

.idea Changes:
* Are there any ways around providing them directly in the repo? Can they be auto generated? I'm personally not a fan of having IDE trace files in the repo as it makes it less digestible and more cluttered.

ninetailedtori:
They can I think! We'd just need to add .idea/ to the gitignore and I think it'll just...autogenerate really :]

Understandable on the others, ill reset then push the gitignore idea ignore and drop the others :]

ninetailedtori:
Moved to pkgbuild-patch branch, due to silly mistakes.

PR #563, opened 2025-12-08
PKGBUILD - patch so it can build.

ninetailedtori:
It seems redundant, now that assets are included within the frontend, and hosted elsewhere, that we don't need to build them externally outside of frontend and sdk, thus we should just scrap the line in the PKGBUILD.

### DIFF
--- a/packages/arch/PKGBUILD
+++ b/packages/arch/PKGBUILD
@@ -2,7 +2,8 @@
 
 pkgver=2.31.0
 pkgname="millennium"
-pkgrel=2
+_pkgdir="Millennium"
+pkgrel=1
 pkgdesc="Millennium is an open-source low-code modding framework to create, manage and use themes/plugins for the desktop Steam Client without any low-level internal interaction or overhead."
 arch=('x86_64')
 url="https://github.com/SteamClientHomebrew/Millennium"


ninetailedtori:
Oddly enough, I've spotted a few other errors. The compilation location seems to be nested in src/millennium_x86-build/ here, and for bootstrap it's in src/millennium_x86-build/boot/.

ninetailedtori:
Currently, this PKGBUILD works perfectly! See my [millennium-git package on the AUR!](https://aur.archlinux.org/packages/millennium-git)

xBiei:
> Currently, this PKGBUILD works perfectly! See my [millennium-git package on the AUR!](https://aur.archlinux.org/packages/millennium-git)

I think you missed copying the assets folder `(/usr/share/millennium/assets)`, I had to do it manually to be able to start steam

ninetailedtori:
Oh darn! Do we need to copy both sdk and frontend or just one? frontend is just for frontend.bin right? Which gets built into the package, I think, so just sdk?

ejalxndr:
`assets` right now only needs `pipx` (`src/frontend`) which is a compat tool for python plugins (we are moving from python to lua, but need a couple buffer releases where both lua and python are supported). But yes, you are correct! Nothing else needs to be in `assets` since sdk AND frontend are statically compiled into the binary now. 

Thanks for the recent help, its much appreciated!

ninetailedtori:
> `assets` right now only needs `pipx` (`src/frontend`) which is a compat tool for python plugins (we are moving from python to lua, but need a couple buffer releases where both lua and python are supported). But yes, you are correct! Nothing else needs to be in `assets` since sdk AND frontend are statically compiled into the binary now.
> 
> Thanks for the recent help, its much appreciated!

Right, that's perfect, I'll push the copy for assets, and then we can merge >:3

And no worries, always happy to help a fun project :3

ninetailedtori:
I FOUND IT, ignore me, lemme push that change now!

xBiei:
> I can't actually find the pipx tool within frontend, do I need to just clone the entire folder over? Or wait, did I misunderstand, has next moved away from pipx? xD

It should be [/src/pipx/](https://github.com/SteamClientHomebrew/Millennium/tree/next/src/pipx)

ninetailedtori:
Noooo wait I just fucked up, I just reset the wrong branch coz I wasn't looking carefully, lemme redo the changes GUH

ninetailedtori:
Sometimes I realise just how much I really just was born for immutable VCS, and it's these moments really

ninetailedtori:
Well, this should be everything, it's completely ok to build now!

ninetailedtori:
Wait, shoot, lemme push some fixes to this LOL. I'll open a new PR, weird naming issues when globbing qwq

Don't revert though, it's cool as is

PR #574, opened 2025-12-20
refactor: singleton rework.

ninetailedtori:
Pretty much:
- Move singleton to its own header.
- Rewrite any singleton classes that didn't inherit, to inherit from singleton.
- Unify get() and getInstance() to GetInstance() which I *think* is code style?

Lemme know if there are any typos, of course, but this flawlessly builds on my machine for all targets. :]


ejalxndr:
Great PR!

ejalxndr:
`housekeep` isn't a standard semantic commit type.

https://www.conventionalcommits.org/en/v1.0.0/

ninetailedtori:
> `housekeep` isn't a standard semantic commit type.
> 
> 
> 
> https://www.conventionalcommits.org/en/v1.0.0/

Fixed

PR #576, opened 2025-12-20
fix(settings): make theme author text be a link only for github

ricewind012:
image


ejalxndr:
If the theme has `theme?.data?.github?.owner` but doesn't have a `theme?.data?.author` nothing will render. We should still fallback onto `theme?.data?.github?.owner`!

ricewind012:
image

PR #581, opened 2025-12-20
Just a cleanup of PKGBUILD.

ninetailedtori:
- Adding back cd (though I don't really like it, we use it throughout anyway, so it's best to stay consistent)
- Move shims to under sdk (ordering sdk actions together, then frontend actions).
- Remove TODO that isn't necessary outside of release, for which I think the release workflow updates the commit tag already for us? If not, that's something we can always add to it.

### DIFF
--- a/packages/arch/PKGBUILD
+++ b/packages/arch/PKGBUILD
@@ -38,7 +38,7 @@ build() {
     echo -e        "\e[1m\e[92m==>\e[0m \e[1mBuilding Millennium...\e[0m"
 
     # Build Millennium!
-    cmake -GNinja  . -DCMAKE_BUILD_TYPE=RelWithDebInfo --preset linux-release -DDISTRO_ARCH=ON
+    cmake -GNinja  . -DCMAKE_BUILD_TYPE=Release --preset linux-release -DDISTRO_ARCH=ON
     cmake --build  build
 }
 


ejalxndr:
Tested?

ninetailedtori:
Tested. See my package for working!

ejalxndr:
We can actually remove `prepare()` entirely as we no longer have any submodules

packages/arch/PKGBUILD
```diff
@@ -17,43 +17,47 @@ sha256sums=('SKIP')
...
+    pnpm --dir     src/frontend      run build
 
-    echo -e    "\e[1m\e[92m==>\e[0m \e[1mBuilding Millennium...\e[0m"
+    echo -e        "\e[1m\e[92m==>\e[0m \e[1mBuilding Millennium...\e[0m"
 
-    cd         $srcdir/$_pkgdir
-    cmake --preset linux-release -G "Ninja" -DDISTRO_ARCH=ON
-    cmake --build build --config Release
+    # Build Millennium!
+    cmake -GNinja  . -DCMAKE_BUILD_TYPE=RelWithDebInfo --preset linux-release -DDISTRO_ARCH=ON
```
shdwmtr: Why `RelWithDebInfo`? For 99% of users, Millennium would just use more disk space with no benefits. If its useful and I'm wrong though, lmk.

PR #604, opened 2025-12-30
feat: add steam url for opening sidebar

ricewind012:
`steam://millennium/sidebar` to open sidebar


ejalxndr:
Nice PR!

We don't actually verify that `context` is `settings` when its not `sidebar`, we just assume it is. This is *fine* as there's only 2 context types, so it will always work as of now; but if we add new options in the future this would break.

ricewind012:
that's what line 62 is, then it returns early for that

ejalxndr:
No, lol. 

> This is fine as there's only 2 context types, so it will always work as of now; but if we add new options in the future this would break.

If we add another option, say "test", then `steam://millennium/settings` & `steam://millennium/test` will both open the settings. All your check does is check if its in the context list, and proceeds if so. However, you only explicitly handle "sidebar" so any other value in the context list will proceed and run L:62 and on. We obviously only want "settings" to do that though, not everything else.

ejalxndr:
@ricewind012 we should get this merged, it's pretty useful.

PR #607, opened 2025-12-30
Fix for ld_preload path.

ninetailedtori:
init_posix.cc : /home/shdw -> std::getenv("HOME")

### DIFF
--- a/scripts/version
+++ b/scripts/version
@@ -1,2 +1,2 @@
 # current version of millennium
-v2.33.0-beta.1
+v2.33.0


ninetailedtori:
Error from: https://github.com/SteamClientHomebrew/Millennium/commit/f448db602774947e848bacebfb1fa0447f4a738a

ejalxndr:
Your clang format messed up some of the shit. 

This isn't a solution though, in a release build we load `libmillennium_hhx64.so` directly from `/usr/lib/millennium/libmillennium_hhx64.so`

In practice though, its always in the same directly as libmillennium, so loading it from the cwd should work.

we should also use a compile time check for debug, and if its debug load it from `./build/.../libmillennium_hhx64.so`
No worries if this makes no sense, you have essentially no context to what libmillennium_hhx64.so even is. I'll fix it if not.

ninetailedtori:
Shoot, hmm. I actually haven't even touched that side yet LOL, you're right, it's an hhx problem.

PR #608, opened 2025-12-31
fix(init-posix.cc): LD_PRELOAD error.

ninetailedtori:


ninetailedtori:
Right @shdwmtr lemme know if I borked that but it *should* work like how we wanted?

ninetailedtori:
couldn't work out if it was meant to use `DEBUG` or `_DEBUG` lmao, used the first but feel free to correct me here hahah.

ninetailedtori:
Can't merge this rn, I'm hitting a segfault:
```
/home/tori/.local/share/Steam/steam.sh: line 928: 901087 Segmentation fault         (core dumped) "$STEAMROOT/$STEAMEXEPATH" "$@"
```

ninetailedtori:
Yep, this is fine to push now, actually it was fine to push at every step of the way, but we're ignoring that.

Issue #609, opened 2025-12-31
[Bug]


MFZealot:
### Before Reporting

- [x] I found no existing issues matching my bug
- [x] My issue is not caused by a theme
- [x] My issue is not caused by a plugin

### Describe the Bug

<img width="633" height="258" alt="Image" src="https://github.com/user-attachments/assets/a912b867-f715-41f5-8b27-890dce98d895" />

Millennium doesn't show the command to update for any reason

### Expected Behavior

Suposed to show the command to update

### Steps To Reproduce

I don't have any ways to reproduce but I'm using Endeavour OS 

### Operating System

Linux

madmaxgrey:
use the same command you used to install millennium.

ejalxndr:
@ninetailedtori we should get this fixed. I don't have time right now but if you do it would be nice to get done.
https://github.com/SteamClientHomebrew/Millennium/blob/main/src/CMakeLists.txt#L175

Someway along the trip from build to the front end, the update command is lost.

ninetailedtori:
Sounds good, lemme fix it up

PR #628, opened 2026-01-07
fix(src/backend/entry_point.cc): Millennium update message showing blank command (#609)

ninetailedtori:
Fixes: #609.

Untested rn, and I need info on whether or not these are the correct strings from NixOS. Would welcome testing on this.

### DIFF
--- a/src/frontend/utils/url-scheme-handler.ts
+++ b/src/frontend/utils/url-scheme-handler.ts
@@ -33,13 +33,67 @@ import { PluginComponent, ThemeItem } from '../types';
 import { Logger } from './Logger';
 import { PyFindAllPlugins, PyFindAllThemes, PyUpdatePluginStatus } from './ffi';
 import { ChangeActiveTheme, UIReloadProps } from '../settings/themes/ThemeComponent';
+import { useQuickAccessStore } from '../quick-access/quickAccessStore';
+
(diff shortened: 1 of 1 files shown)


ejalxndr:
We should do it at compile time in cmake. We already have it setup that way. This reduces startup overhead and incorrect detections. For example, if I was on arch, but manually installed (i.e outside of the aur) I would be prompted to use the aur command to update, even though I installed manually. 

I'll take a look at it in a couple days likely if we don't get it resolved in the meantime.

ninetailedtori:
> We should do it at compile time in cmake. We already have it setup that way. This reduces startup overhead and incorrect detections. For example, if I was on arch, but manually installed (i.e outside of the aur) I would be prompted to use the aur command to update, even though I installed manually.
> 
> I'll take a look at it in a couple days likely if we don't get it resolved in the meantime.

Done.

src/frontend/index.tsx
```diff
@@ -86,8 +86,9 @@ async function initializeMillennium(settings: SettingsProps) {
 
 	patchMissedDocuments();
 
-	const notificationService = new NotificationService();
-	notificationService.showNotifications();
+	let showNotifications: () => Promise<void>;
+	({showNotifications} = new NotificationService());
+	await showNotifications();
```
shdwmtr: ?

PR #630, opened 2026-01-08
Refine Russian localization strings

authorless:
Updated Russian localization strings for clarity and consistency.


authorless:
@shdwmtr I have corrected the Russian localization and added missing elements that were not there before. Could you please check it? If everything is correct, we can merge these changes. I've thoroughly double-checked them, so there shouldn't be any issues.

dotFelixan:
Отличный перевод!
Спасибо что исправил мои косяки😭

PR #634, opened 2026-01-15
Make Updater weak_ptr and add const/style fixes

ninetailedtori:
Replace shared_ptr<Updater> with weak_ptr and guard progress dispatches with .lock(). Mark several APIs const or static and tighten parameter types. Reorder headers, prefer free filesystem helpers (exists/create_directories), and clean up minor style issues. Fix plugin install path handling and correct inline documentation in theme_mgr.h.

### DIFF
--- a/src/backend/plugin_mgr.cc
+++ b/src/backend/plugin_mgr.cc
@@ -28,18 +28,17 @@
  * SOFTWARE.
  */
 
+#include "head/plugin_mgr.h"
 #include "head/entry_point.h"
 #include "head/library_updater.h"
-#include "head/plugin_mgr.h"
 #include "head/scan.h"
 
-#include "millennium/logger.h"
-#include "millennium/http.h"
-#include "millennium/env.h"
 #include "millennium/encode.h"
+#include "millennium/http.h"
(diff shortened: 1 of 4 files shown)


src/backend/plugin_mgr.cc
```diff
@@ -208,30 +235,30 @@ bool plugin_installer::update_plugin(const std::string& id, const std::string& n
...
-            std::filesystem::remove_all(tempDir);
+            remove_all(tempDir);
             return false;
         }
 
-        std::filesystem::path targetFolder = get_plugins_path() / name;
+        const std::filesystem::path targetFolder = get_plugins_path() / name;
         Logger.Log("Preparing target plugin directory: {}", targetFolder.string());
-        std::filesystem::create_directories(targetFolder);
+        create_directories(targetFolder);
```
shdwmtr: std::filesystem is better here (and for other occurrences), its more explicit, and less likely to accidentally reference a function that isn't apart of the filesystem namespace.

ejalxndr:
Thanks for the PR!

shdwmtr:
Fixed in `next`.

PR #635, opened 2026-01-16
Fix steamwebhelper crash when opening Chrome Web Store

Ingan121:
When you open https://chromewebstore.google.com/ in the Steam browser with Millennium loaded, steamwebhelper.exe immediately crashes. This PR resolves the issue by excluding the page from the HTTP hooks.
To test the issue, drag and drop the above link to the store page in the main window, or open that URL in the overlay browser. I confirmed this behavior on Windows 10 and 11.

After tinkering with the code, I found that sending the `Fetch.fulfillRequest` message causes the crash. I'm guessing that this is caused by internal conflict with the DevTools Fetch API and the internal Chrome extension that enables private API access for the web store (`chrome-extension://ahfgeienlihckogmohjhadlkjgocpleb`).

### DIFF
--- a/src/core/http_hooks.cc
+++ b/src/core/http_hooks.cc
@@ -177,7 +177,10 @@ static const std::unordered_set<std::string> g_doNotHook = {
 
     /** Ignore youtube related content */
     R"(https?://(?:[\w-]+\.)*(?:youtube(?:-nocookie)?|youtu|ytimg|googlevideo|googleusercontent|studioyoutube)\.com/[^\s"']*)",
-    R"(https?://(?:[\w-]+\.)*youtu\.be/[^\s"']*)"
+    R"(https?://(?:[\w-]+\.)*youtu\.be/[^\s"']*)",
+
+    /** Ignore Chrome Web Store (causes a webhelper crash on Fetch.fulfillRequest) */
+    R"(https?:\/\/(?:[\w-]+\.)*chromewebstore\.google\.com\/[^\s"']*)",
 };
 // clang-format on
 


ejalxndr:
Thanks for the PR! The web store isn't usable regardless though, right? I've never even tried going to that URL as I assumed Steam would have locked it down.

Ingan121:
image

It actually works, as Valve enabled the CEF's Chrome Runtime feature around mid 2025. You can manage extensions in `chrome://extensions` as well.

A file dialog appears when installing extensions on the web store. You can save it anywhere, as it'll be removed once the installation finishes.

ejalxndr:
Wow, that changes a lot. Where are the patch notes for this change, do you have them? I remember trying this before to no avail many many years ago.  

Its likely possible to programmatically wrap their extension API to make extensions installable from Millennium directly, without having to mess with CEF - this would be a cool feature!

Ingan121:
I discovered this change in November 2025, and AFAIK, it wasn't active in early to mid 2025. There seems to be no specific mention of 'enabled Chrome runtime' in the official Steam changelogs, although some entries in the November and December ones imply Chrome runtime being enabled.

[Chrome runtime is mandated since CEF 128](https://groups.google.com/g/cef-announce/c/s1WaovAopFo/m/LV5eiNX1BgAJStarting), thus I think Valve enabled it in preparation for bumping the CEF version. (It's still stuck in 126, though.)

With Chrome runtime enabled, it is also possible to pop open the full Chromium browser UI out of steamwebhelper.exe. There are multiple ways to do so, and one of the ways is running the following code in SharedJSContext: `await MILLENNIUM_API.ChromeDevToolsProtocol.send("Target.createTarget", {url:"chrome://extensions"})`

Programmatically installing extensions:
Installation of signed extensions from the web store could be initiated by running JS in the web store context, but I have no idea how to bypass the "install extension?" pop-up. (File dialog can be skipped by running the web store page in a CEF window unmanaged by Steam code, e.g., the full browser UI.) Unsigned extensions are more complicated; there are related APIs in `chrome://extensions` (`chrome.developerPrivate`), but none of them support specifying arbitrary file paths.
The `--load-extension` argument exists and is likely to work if it's provided to steamwebhelper.exe, and also [Extensions.loadUnpacked](https://chromedevtools.github.io/devtools-protocol/tot/Extensions/#method-loadUnpacked) DevTools API might work. (Doesn't seem to work with `MILLENNIUM_API.ChromeDevToolsProtocol`, though.)

Querying installed and enabled extensions is possible by calling `await chrome.developerPrivate.getExtensionsInfo()` in `chrome://extensions`.

Ingan121:
Made an experimental frontend plugin for managing and accessing Chrome extensions in the Steam client: [ExtManager](https://github.com/Ingan121/ExtManager)

Maybe this helps develop the integrated extension manager.

ejalxndr:
Crap, sorry about that; I should have updated you. We are essentially adding an extension manager directly into Millennium. We're working on it right now in the `async-ffi` branch, feel free to contribute!

On that note, your plugin might not useful considering these changes. Sorry for the wasted time.

PR #640, opened 2026-01-22
replace deprecated prev.system

powerofthe69:
system as referenced in `inherit (self.packages.${prev.system}`, is deprecated which results in these warnings when rebuilding:
`evaluation warning: 'system' has been renamed to/replaced by 'stdenv.hostPlatform.system'`

This renames it to the proper name which is now `stdenv.hostPlatform.system` inside the `flake.nix`. Testing on my local appears to have made the error go away.

Other changes were due to my `prettierd` adjusting the formatting to be in line with the `nixfmt` standard. If this is unwanted, then it can be reverted, but I think it makes it more legible than it was before

### DIFF
--- a/packages/nix/flake.nix
+++ b/packages/nix/flake.nix
@@ -52,19 +52,39 @@
     {
       packages.x86_64-linux =
         let
-          pkgs = import nixpkgs { system = "x86_64-linux"; config.allowUnfree = true; };
+          pkgs = import nixpkgs {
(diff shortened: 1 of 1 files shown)


ejalxndr:
@Trivaris

Trivaris:
Yep looks good to me

Trivaris:
Sorry for the late reply, didn't see it until now

PR #642, opened 2026-01-26
Update schinese.json with missing translations

WittF:
- Add 23 missing translation keys to sync with english.json
- Use "快捷 CSS" instead of "快速 CSS" for Quick CSS feature


WittF:
btw, I noticed many localization keys in the language files are no longer used in the codebase, is this intentional or should they be cleaned up?

ejalxndr:
Yeah, they should be - it's not intentional I'm just lazy

PR #644, opened 2026-01-29
Improve i18n coverage and clean up unused locale keys

WittF:
## Summary                                                                                    
  - Replace hardcoded strings in settings UI components with locale references                  
  - Import formatString utility for parameterized translations                                  
  - Remove unused locale keys across all language files                                         
                                                                                                
  ## Changes                                                                                    
  - Updated settings components (plugins, themes, logs, updates, quickcss) to use locale keys   
  - Added complete translations for English and Simplified Chinese                              
  - Removed deprecated locale keys from all language files


ejalxndr:
Thanks, I've been meaning to fix those locale issues! To be honest, at this point maybe we should consider actually adding [React i18n](https://www.npmjs.com/package/react-i18next)?

WittF:
> Thanks, I've been meaning to fix those locale issues! To be honest, at this point maybe we should consider actually adding [React i18n](https://www.npmjs.com/package/react-i18next)?

Looked into `react-i18next` - its good, but might be overkill for our scale (168 keys, 24 components). Current solution is 120 lines with zero deps, react-i18next adds ~20kB.

@shdwmtr I think the current approach works well enough, what do you think?

ejalxndr:
Yeah, good point - let's stick with the current approach. One downside to the current approach is that new contributors won't easily know which locales are active, deprecated, or unused. It's also just cluttered to begin with.

ejalxndr:
Maybe we can somehow make it easier to configure locales for Millennium. Maybe a web app or something. I'm pretty sure there are some options out there but I'm not sure if there are paid. They also rely on React i18n.

WittF:
> Maybe we can somehow make it easier to configure locales for Millennium. Maybe a web app or something. I'm pretty sure there are some options out there but I'm not sure if there are paid. They also rely on React i18n.

Like https://crowdin.com/? It's free for OSS. Maybe we can try it, though not sure if that'd overcomplicate things for us.

WittF:
Alternatively, we could start with a few custom scripts (locale:check, locale:fix, locale:types) that scan code for `locale.xxx` patterns and generate TS types from english.json for active keys only, enabling basic key analysis and cleanup 🤔

ejalxndr:
Hmm, crowdin looks interesting. I'll look into that.

ejalxndr:
I cherry-picked about 90% of your commits. I didn't merge this PR directly since it's diverged fairly far from main and rebasing would have been a lot of work.

Feel free to open another PR if there are any further changes needing to be made. 
Cheers!

PR #648, opened 2026-02-07
Prepare Nix building for nixpkgs integration

Trivaris:
The current implementation of nix packaging built its own python version (3.11.8), as only 3.11.14 is available within nixpkgs.
This comes with a few drawbacks, the biggest being that nixpkgs does not want us to do this.
So in preparation to the merge in [#487045](https://github.com/NixOS/nixpkgs/pull/487045) I removed the hardcoding of the python library version - only in Nix builds of  course - by adding the NIX_PYTHON_LIB flag.
It is only used in `src/CMakeLists.txt` within blocks that only get run if `DISTRO_NIX` is set, as millennium already perfecly handles everything else dynamically.

There are also a few changes to the nix packages:
- Millennium x86 and x64 are now being built separatly, as nixpkgs needs the meta attribute, which specifies the architechture that the package is being built in. To avoid errors regarding unsupported arch, I used this approach.
- Small fix regarding OpenSSL in millennium-steam
- As mentioned above, the millennium-python has been removed entirely
- While I generally agree with nixfmt formatting, for the packages output I chose to format it manually to make it easier to understand what is going on

### DIFF
--- a/packages/nix/flake.lock
+++ b/packages/nix/flake.lock
@@ -158,13 +158,13 @@
       "locked": {
         "lastModified": 1770463863,
         "narHash": "sha256-MceGTpXobCAh5Ll/1iTWcEf6/nHY3Ll1t06JxbYc+Co=",
-        "owner": "Trivaris",
+        "owner": "SteamClientHomebrew",
         "repo": "Millennium",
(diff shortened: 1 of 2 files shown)


ejalxndr:
Let me know when this is good to merge

Trivaris:
Should be good now :+1:

PR #671, opened 2026-03-15
fix(theme): prevent repeated update prompt after theme update and restart

### LINKED ISSUE #649: [Bug] I am prompted to update theme and restart everytime !
### Before Reporting

- [x] I found no existing issues matching my bug
- [x] My issue is not caused by a theme
- [x] My issue is not caused by a plugin

### Describe the Bug

 am prompted to update theme and restart everytime ! and it restarted and always ask

### Expected Behavior

should update

### Steps To Reproduce

wait an update, download, restart. It will prompt again

### Operating System

Windows

### Anything else?

_No response_

BillionClaw:
## Description

This PR fixes the issue where users are prompted to update a theme and restart Steam, but after restarting, they are prompted again for the same theme update.

## Root Cause

Two issues were identified:

1. **Silent git fetch failures**: The `git_remote_fetch` function in `UpdateTheme()` didn't check its return value, so if the fetch failed, the reset would use stale remote HEAD data.

2. **Cached update state**: After a theme update, the cached update check state wasn't being properly reset, which could cause stale data to persist.

### Backend (C++):
- Added `ResetCheckState()` method to clear cached updates after a theme update
- Added error handling for `git_remote_fetch` to prevent silent failures

### Frontend (TypeScript):
- Added FFI function to expose reset functionality
- Theme update now resets the check state after successful update

## Testing

After this fix:
- Theme updates should properly clear the cache
- On next Steam restart, fresh update data will be fetched
- Git fetch failures are now properly detected

Fixes #649

### DIFF
--- a/src/backend/entry_point.cc
+++ b/src/backend/entry_point.cc
@@ -135,6 +135,7 @@ IPC_RET(Core_SetBackendConfig, CONFIG.SetAll(nlohmann::json::parse(ARGS["config"
 
 /** Theme and Plugin update API */
 IPC_RET(Core_GetUpdates, updater->CheckForUpdates(ARGS.value("force", false)).value_or(nullptr))
+IPC_NIL(Core_ResetUpdateCheckState, updater->ResetCheckState())
 
 /** Theme manager API */
 IPC_RET(Core_GetActiveTheme, themeConfig->GetActiveTheme())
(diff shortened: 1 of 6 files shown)


ejalxndr:
Should be fixed in the latest beta release.

BillionClaw:
Thanks for confirming! Closing as this appears to be fixed in the latest beta release.

PR #684, opened 2026-03-20
fix: resolve Linux shutdown hang caused by reconnect loop

powerofthe69:
On Linux, when Steam exits, the frontend WebSocket disconnects but StartFrontEnds() immediately recurses to reconnect — before StopMillennium() has a chance to set the termination flag. This creates a deadlock where the process never exits.

This change:
- Extends the 10-second reconnect wait (previously Windows-only) to all platforms, checking g_shouldTerminateMillennium during the window
- Calls BackendManager::Shutdown() from EntryMain() on all platforms, not just Windows, so plugin _unload() hooks are invoked reliably
- Adds a guard flag to prevent double-shutdown from the destructor

Please review - I do not know C that well so I leveraged Claude to implement this. I wasn't sure how the project felt about LLM-assisted changes, so if there's a more efficient way, or a reason why this wasn't supported up til now then please make changes or suggestions!

I just wanted to bring this to attention. Steam was not exiting due to _unload hooks not being called, allowing child processes created by plugins to reside in the background.

### DIFF
--- a/src/core/backend_mgr.cc
+++ b/src/core/backend_mgr.cc
@@ -210,6 +210,9 @@ done:
  */
 void BackendManager::Shutdown()
 {
+    if (m_hasShutdown) return;
+    m_hasShutdown = true;
+
     Logger.Warn("Unloading {} plugin(s) and preparing for exit...", this->m_pythonInstances.size() + this->m_luaThreadPool.size());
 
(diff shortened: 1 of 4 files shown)


shdwmtr:
Use the latest beta, should fix the issue.

PR #687, opened 2026-03-21
feat(next):Clean PKGBUILD for next branch

ninetailedtori:


shdwmtr:
The url should stay as the repositories source. The only useful change here might be adding `prepare()` but there's no huge reason to do that.

ninetailedtori:
> The url should stay as the repositories source. The only useful change here might be adding `prepare()` but there's no huge reason to do that.

Hmm fairs :]

PR #693, opened 2026-03-24
feat: allow passing --beta argument for installing beta prereleases

CatPieLeaf:
### Now it's possible to install beta prereleases on distributions other than arch and nix by passing `--beta` argument to install.sh


shdwmtr:
Been meaning to get to this for a while, thanks!

PR #716, opened 2026-04-19
fix(nix): Fixed build bug with python venv

cowburrs:
Millennium on nix doesn't work out of the box without this build environment variable. venv needs the file location of python. will send images.

I probably should have created a issue first


cowburrs:
``` 
Traceback (most recent call last):
  File "/usr/share/millennium/assets/pipx/main.py", line 40, in <module>
    config = config.Config()
             ^^^^^^^^^^^^^^^
  File "/usr/share/millennium/assets/pipx/config.py", line 37, in __init__
    self.setup()
  File "/usr/share/millennium/assets/pipx/config.py", line 62, in setup
    os.chmod(self.PYTHON_BIN, 0o755)
FileNotFoundError: [Errno 2] No such file or directory: '/home/burrs/.local/share/millennium/.venv/bin/python3.11'
```

One of the errors I was getting.

```
╰─$ steam
[Millennium Steam] venv not found. Millennium will set it up on next launch
```

The other error I was getting.

cowburrs:
> ```
> ╰─$ steam
> [Millennium Steam] venv not found. Millennium will set it up on next launch
> ```

I should mention that multiple restarts did not solve the issue.

cowburrs:
after a little bit of further investigation I think I know what the core issue with my system was

`> curl -fsSL "https://steambrew.app/install.sh" | bash`

I had originally run this command, which is incompatible with nixos, and it had likely left some sort of python install in my venv

just adding this environment variable circumvents that

cowburrs:
yeah ok that is the problem

so there are reproduction steps to this bug

delete the millenium venv (if you have it)
`> rm -rf $HOME/.local/share/millennium/.venv/`
run the following generic installation command
`> curl -fsSL "https://steambrew.app/install.sh" | bash`
then attempt to run millennium-steam
`> nix run github:SteamClientHomebrew/Millennium?dir=packages/nix#millennium-steam`

so a very niche fix xD

PR #727, opened 2026-05-03
removing 'shallow=true' attributes from inputs url as it is irrelevant and breaking for lix

Murazaki:
Several users have been having issues importing the nix configuration because of the `shallow=true` flag, because they use Lix in place of Nix.

I discussed this directly with some Lix/Nix developers and they stated that the `shallow` flag does not make sense here, as  github flakerefs are fetched as commit tarballs, and thus would not fetch the git tree.

This is why I'm proposing to get rid of the flag, as it is considered irrelevant from reputable sources.

### DIFF
--- a/packages/nix/flake.nix
+++ b/packages/nix/flake.nix
@@ -4,22 +4,22 @@
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-unstable";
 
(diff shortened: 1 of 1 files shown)


shdwmtr:
@Trivaris

Trivaris:
Looks good to me 👍

PR #748, opened 2026-05-15
Edited dependencies in millennium(and 32).nix and migrated nix build script to bun

kkuroma:
Fixed `pkgsi686Linux` and `cacert` unresolved variable issues in the nix files by referencing `pkgs`. Since the main repository moved to bun over pnpm, the nix build now creates a bun-built frontend.

Noteworthy changes/caveats
1.  removed all references to `pnpm`
2. the `sed` in millennium-32.nix is patching over an upstream bug (`#define UPDATER_DEBUG_MODE` left in a release build path)
3. `-Wno-error=c++26-extensions` patch, since GCC allows `#embed` as an extension in `C++23` mode but `-Wpedantic -Werror` rejects it
4. `pvs64` is now compiled in millennium-64 rather than the 32-bit build, since `pkgsi686Linux.stdenv` cannot emit 64-bit code.
5. `cdp-isolated-ctx` was missing from the shims build and has been added
6. assets.nix used to refer to `src/pipx` which was removed in `bf091385`
7. Since nixpkgs bumped the version of bun from 1.3.8 to 1.3.13, flake.lock was updated. The bun FOD hashes in frontend.nix and shims.nix were recomputed accordingly.
8. minhook, mini, libgit2 inputs removed from flake.nix as they are no longer referenced in the source.

The build successfully compiled and the resulting steam client worked visually.

### DIFF
--- a/packages/nix/assets.nix
+++ b/packages/nix/assets.nix
@@ -15,7 +15,6 @@ stdenv.mkDerivation (finalAttrs: {
     runHook preInstall
 
     mkdir -p $out/share/millennium/assets/
-    cp -r src/pipx $out/share/millennium/assets/pipx
 
     runHook postInstall
   '';
--- a/packages/nix/flake.lock
+++ b/packages/nix/flake.lock
@@ -105,17 +105,16 @@
     "millennium-src": {
       "flake": false,
       "locked": {
-        "lastModified": 1777483535,
-        "narHash": "sha256-eqbNqc+5PibHLI8xgOrizXOVLcpE52MhmUR4o40S3PY=",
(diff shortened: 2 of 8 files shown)


flyingpeakock:
Have you tried to fix the Bun dependency on the "next" branch? The nix files have been greatly simplified there and seems like it would be much easier to maintain than the current way this package is built for nix.

kkuroma:
I might attempt it there if I have time. I agree that the next branch is much more organized and more easily maintained than specific compilation flag hacks.

ejalxndr:
Fixed in #551

PR #758, opened 2026-05-23
fix(locales): evaluate settings tab titles at render time

WkdXeqtr:
## Problem

Tab objects in `settings/index.tsx` were defined at module level, causing their `title` properties to be evaluated before `GetLocalization()` resolved the user's language. As a result, sidebar tab names ("General", "Themes", "Quick CSS", etc.) always appeared in English regardless of the selected language.

## Fix

Moved tab definitions inside the `MillenniumSettings` component so titles are read from `locale` at render time, after localization has loaded.

### DIFF
--- a/src/typescript/frontend/settings/index.tsx
+++ b/src/typescript/frontend/settings/index.tsx
@@ -50,78 +50,6 @@ declare global {
 	const g_PopupManager: any;
 }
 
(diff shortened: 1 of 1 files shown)


shdwmtr (approved):

ejalxndr:
Open the same PR against the `next` branch if you want these changes in the next beta release.

PR #762, opened 2026-05-25
fix: OpenSSL 3.5/GCC 16 compatibility + Bootstrap arch fix

EliasOulkadi:
### 1. Bootstrap `-m32` (src/boot/linux/CMakeLists.txt)
The bootstrap target uses a nested `project()` call which doesnt inherit `CMAKE_C_FLAGS` from the parent. Added explicit `-m32` via `target_compile_options` / `target_link_options`. Otherwise the 32-bit Steam process cannot load a 64-bit bootstrap `.so`.

### 2. pnpm workspaces (src/sdk/package.json)
Added `"workspaces": ["packages/*"]` so pnpm resolves linked packages (`@steambrew/client`, `@steambrew/webkit`) correctly.

### 3. LuaJIT CCallState fix (build/_deps/luajit-src/src/lj_ccall.h)
On GCC 16+ with `LJ_TARGET_X86`, the `CCallState` struct lacks `nfpr` member but `vm_x86.dasc` references it. Patch:
```c
#elif LJ_TARGET_X86
  uint8_t nfpr;          /* Number of arguments in FPRs. */
  uint8_t resx87;        /* Result on x87 stack: 1:float, 2:double. */
```

### 4. OpenSSL 3.5 ENGINE removal
OpenSSL 3.5 removed the ENGINE API. Build with `-DOPENSSL_NO_ENGINE` to avoid compile errors.

## Build tested on
- Fedora 44, GCC 16.1, kernel 6.x
- 32-bit main lib ✓, 64-bit hook helper ✓, 32-bit bootstrap ✓

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -161,7 +161,11 @@ jobs:
             - name: Build Millennium Backend
               shell: pwsh
               run: |
-                  pushd "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\Tools"
+                  $vsInstallPath = & "C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" `
+                    -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 `
+                    -property installationPath
+                  $vsToolsPath = Join-Path $vsInstallPath "Common7\Tools"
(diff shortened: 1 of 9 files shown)


ejalxndr:
These changes have already landed, this PR has many merge conflicts. I'll just close this as its no longer useful.

PR #774, opened 2026-05-30
fix(locales): localize plugin installer status messages

WkdXeqtr:
## Problem

Status messages shown in the plugin installation progress dialog were
hardcoded in English inside the C++ backend (`plugin_mgr.cc`) and
passed as raw strings to the TypeScript frontend via `dispatch_progress()`.

## Fix

- Replace hardcoded strings in `plugin_mgr.cc` with locale key names
- Add `resolveStatus()` helper in `InstallerProgress.tsx` that looks up
  each key in the localization system, falling back to the raw string
  if the key is not found (safe for future additions)
- Add locale keys to `english.json` with the original English text
- Add Russian and Ukrainian translations


ejalxndr:
we should probably explicitly set a "##" prefix or something to denote a locale vs real user data. Something like "##localeStr" is usually preferable!

WkdXeqtr:
Sure, sorry about that - I wasn't aware of the convention. Updated!

WkdXeqtr:
Sorry, I just noticed you changed the base branch from next to main - did I do something wrong? I'm new to GitHub PRs. I was planning to target all my changes at the next branch, did I submit from the wrong one?

ejalxndr:
`next` is the perfect target! For this specific case I merged into main to get this out quicker.

WkdXeqtr:
Sorry for the late follow-up - won't this cause parallel development issues if some changes are in main and some are in next? I assumed the workflow was linear - new changes go into next, then next replaces main at some point, and so on.

ejalxndr:
@WkdXeqtr Nope! This is how its done professionally as well! I can merge histories, cherry-pick, and fast-forward either branch incredibly easy. That's what those Git features were created for!

I'd recommend always targeting `next` assuming I want it there. If I need it somewhere else I'll move it.

PR #777, opened 2026-05-31
fix: defer tier0 hook installation to avoid loader lock crash

5bz5928:
## Problem
Steam build 1779918128 (May 2026) introduced a `CFileWriterThread` that conflicts with Millennium's `DllMain` initialization. `register_dll_notifications()` calls `handle_already_loaded()` → `dll_notification_callback()` → `handle_tier0_dll()`, which calls `VirtualProtect` and `GetProcAddress` — all from inside the Windows loader lock. This causes a deadlock/crash (STATUS_ABANDONED, `std::terminate` in `CFileWriterThread::Run`).

## Fix
Defer tier0 hook installation from `register_dll_notifications()` (called from `DllMain` while inside the loader lock) to `initialize_steam_hooks()` (called from the Millennium background thread, outside the loader lock).

A `std::vector<PVOID>` with a mutex queues tier0 module handles during `DllMain`. They are processed later by `process_deferred_tier0_hooks()` in `initialize_steam_hooks()`.

## Testing
- ✅ Steam starts without crashing (loader lock crash eliminated)
- ✅ Millennium crash handler + background thread initialize safely
- ✅ Steam UI loads; web helper spawning path is hooked

## Notes
This affects Windows only (the Linux/macOS paths are unchanged). The queue/process pattern is thread-safe: the mutex protects against concurrent access from both the DLL notification callback and the background thread.

### DIFF
--- a/src/instrumentation/internal/steam_hooks.cc
+++ b/src/instrumentation/internal/steam_hooks.cc
@@ -33,6 +33,7 @@
 #include <mutex>
 #include <string>
 #include <string_view>
+#include <vector>
 
 #ifdef __linux__
 #include <dlfcn.h>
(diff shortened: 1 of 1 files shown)


ejalxndr:
`CFileWriterThread` has always existed. The hooks are intentionally setup while holding the loader lock, its not a bug. I can't reproduce your crash either, and no one has reported a similar error.

PR #787, opened 2026-06-04
feat(quick-access): add floating quick access button

WkdXeqtr:
## Summary

On Steam Deck the sidebar has a dedicated hardware button. On desktop
you have to go through the Steam menu - which gets old fast when
you regularly check extensions.

Adds a floating semi-transparent button on the right edge of the window.
Half the circle stays visible at all times (clipped by viewport or panel),
slides in sync with the sidebar animation, and toggles it on click.

## Settings

New section in Millennium Settings -> General -> **Quick Access Button**:
show/hide toggle + location dropdown (Library only / Library except game
pages / Everywhere). Default: Library except game pages. Both settings
apply instantly without a reload.

## Performance

- Zustand slice: sidebar re-renders only on QAB state changes
- Navigation via `pushState`/`replaceState` patches, not polling
- 150 ms debounce to avoid flicker on Steam's multi-step navigation
- 1 s startup delay for `MainWindowBrowserManager.m_lastLocation`


ejalxndr:
Logic is rather hacky. I don't have a better solution either, but this definitely isn't it. Regardless, the project isn't in a state where we are accepting new UI features like this - we just got a release out and we are focused on stability. I could see this option in the future, but not now. 

Thank you for you hard work!

PR #793, opened 2026-06-06
feat(quick-access): expose window.MILLENNIUM_SIDEBAR for plugins

WkdXeqtr:
### Summary

Adds `MillenniumSidebar` to `@steambrew/client` - a small, typed API that lets a
plugin drive the desktop sidebar (the quick-access panel that hosts plugin
settings):

- `MillenniumSidebar.openPlugin(name): Promise<boolean>` - open the sidebar
  focused on a plugin by its internal name (same as the plugin manager's
  "Configure"). Resolves `false` if no such plugin.
- `MillenniumSidebar.open(): void` - open the sidebar on its home (plugin/theme list).
- `MillenniumSidebar.close(): void` - close the sidebar.

The implementation lives in Millennium's frontend (it owns the sidebar store), so
the frontend registers it with the SDK when the sidebar mounts; plugins only call
the public methods. No global is exposed to plugins.

### Details

A plugin can render a panel in the sidebar, but nothing lets it *open* that panel
from its own code - only the user clicking "Configure" can. I'm building a plugin
that shows a toast and, when you click it, opens its panel in the sidebar; that
isn't possible today, and this PR makes it possible.

This is also groundwork for plugin interop / dependencies (asked for in issues).
With it a plugin can:

- open **its own** settings on an action - e.g. a notification's onClick, or a
  first-run "finish setup" prompt;
- open **another** plugin's settings - useful for compatibility / "this needs
  plugin X setting" flows;
- open or close the sidebar programmatically.

### Usage

```ts
import { MillenniumSidebar } from '@steambrew/client';

MillenniumSidebar.openPlugin('my-plugin');    // open my own panel
MillenniumSidebar.openPlugin('other-plugin'); // open another plugin's panel
MillenniumSidebar.open();                      // open the list
MillenniumSidebar.close();                     // close the sidebar
```

### DIFF
--- a/src/typescript/frontend/quick-access/index.tsx
+++ b/src/typescript/frontend/quick-access/index.tsx
@@ -28,7 +28,7 @@
  * SOFTWARE.
  */
 
-import { ErrorBoundary, getParentWindow } from '@steambrew/client';
+import { ErrorBoundary, getParentWindow, registerMillenniumSidebar } from '@steambrew/client';
 import { useCallback, useEffect, useRef, useState } from 'react';
 import { TitleView } from '../components/QuickAccessTitle';
 import { PluginSelectorView, RenderPluginView } from './PluginView';
@@ -36,6 +36,7 @@ import { BrowserManagerHook } from './browserHook';
 import Styles, { MillenniumDesktopSidebarStyles } from '../utils/styles';
 import { DesktopSideBarFocusedItemType, useDesktopMenu } from './DesktopMenuContext';
 import { useQuickAccessStore } from './quickAccessStore';
+import { backend } from '../utils/ffi';
 import { RenderThemeView, ThemeSelectorView } from './ThemeView';
 import { Placeholder } from '../components/Placeholder';
 import { MillenniumIcons } from '../components/Icons';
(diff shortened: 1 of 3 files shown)


shdwmtr (changes requested):
Plugins should never have raw access to an API, they should program to an abstraction. Add a API to `@steambrew/client` (its at `src/typescript/sdk`).

WkdXeqtr:
How about this? Moved it to a typed `MillenniumSidebar` API in `@steambrew/client` (`src/typescript/sdk/packages/client/src/millennium-sidebar.ts`). The frontend registers the implementation when the sidebar mounts (`registerMillenniumSidebar`, marked `@internal`), plugins just call `MillenniumSidebar.openPlugin/open/close`, and no global is exposed. Updated the PR.

Btw, do my other PRs look OK, or is something off with them? #787, and SteamClientHomebrew/Installer#10.

And I also wanted to try
- a URL scheme to install plugins/themes straight from the website (an in-client "Install" deep link);
- a localization system for the website;
- a localization system for plugins - either baked into the plugin template, or provided via a dependency on a shared plugin (which is part of what this sidebar API unlocks).

What do you think?

ejalxndr:
Hello! I haven't gotten to your other PR's yet. 

More on this though: plugins should not be allowed to install new content with or without permission.
As for locales, that could be a good idea, but it would need proper planning. 

I've made a commit in [`next`](https://github.com/SteamClientHomebrew/Millennium/commit/d5fee0ccf38818eea85b09cd4e6229204af71ea9) that adds support for plugins to open quick access. Easier for me to implement it then going back and forth! It will be in the next beta release.

WkdXeqtr:
Thanks, that works for me.
What did you mean by "plugins should not be allowed to install new content"?
I was actually referring to Millennium itself. You already have the steam/millennium schema. However, installing plugins currently requires copying the ID from the website, which is inconvenient. What I was suggesting is to create a schema that skips the copying step and instead triggers an installation prompt directly from a button on your site. This would be handled in Millennium, rather than in the plugin.

ejalxndr:
We actually used to have that. Over the years of enumerable issues, we removed it. The current method is much simpler and easier to maintain. It's exactly what it should be; nothing more, nothing less.

PR #806, opened 2026-06-13
Refactor WebKit URL validation to use libcurl

BlythT:
To prevent https://github.com/SteamClientHomebrew/Millennium/pull/803 happening again, I refactored the URL validation:
- Extracted validation logic into a class `target_url`
- Replaced the existing regex and string matching with libcurl (already imported so no added dependencies) for more robust URL parsing
- Added exhaustive testing to detect automatically if this breaks in future

I wanted to add ` ctest --test-dir build --output-on-failure -C Release` to the github actions ci.yml but I encountered unrelated test failures which would block build if merged https://github.com/BlythT/Millennium/actions/runs/27465672305 so I've reverted that


BlythT:
Rebased this

PR #817, opened 2026-06-23
fix(arch): remove hardcoded `debug` option

gwuen:
The [PKGBUILD reference](https://wiki.archlinux.org/title/PKGBUILD#options) describes the `options` array as a way of “overriding some of the default behavior of *makepkg*, defined in `/etc/makepkg.conf`”. This means that setting `debug` here forces debug package creation even for users who have explicitly opted out via `!debug` in their global config.

A package should not silently override a deliberate user preference without a compelling technical reason. None exists here; `makepkg` handles debug packages correctly on its own when the user wants them.

---

I'd also like to kindly ask that you monitor the AUR comment section. The [AUR Submission Guidelines](https://wiki.archlinux.org/title/AUR_submission_guidelines#Maintaining_packages) explicitly state that maintainers should “check for feedback and comments from other users and try to incorporate any improvements they suggest”. Redirecting all discussion to GitHub excludes users without an account.

I raised this concern in the AUR comments on 2026-04-21 and again on 2026-06-08 with no response. Additionally, the [submission rules](https://wiki.archlinux.org/title/AUR_submission_guidelines#Rules_of_submission) ask that the maintainer header include a contact email address, which is currently missing from this PKGBUILD.

### DIFF
--- a/packages/arch/PKGBUILD
+++ b/packages/arch/PKGBUILD
@@ -1,4 +1,4 @@
-# Maintainers: SteamClientHomebrew <https://github.com/SteamClientHomebrew>
+# Maintainer: SteamClientHomebrew <noreply@steambrew.app>
 
 pkgname="millennium"
 pkgver=3.3.0
--- a/packages/arch/PKGBUILD
+++ b/packages/arch/PKGBUILD
@@ -12,7 +12,6 @@ makedepends=('git' 'bun' 'curl' 'zip' 'unzip' 'tar' 'cmake' 'ninja' 'lib32-gcc-l
 install=millennium.install
 source=("git+$url.git#commit=948d3bb57f811d4bd9a1f43963defe73253a05bf")
 sha256sums=('SKIP')
-options=(debug)
 
 _pkgdir="Millennium"
 


shdwmtr (approved):

ejalxndr:
Thanks for the PR, I also fixed the contact section. As for the checking the AUR comment section, I either forget, or don't have time to look at it. This project is already a lot to maintain, let alone the AUR port. Many other FOSS do the same. As a general rule of thumb, I will not go out of my way to look for those in need of support - they should come to me; I only check GitHub for issues, direct email, or issues directly submitted by community moderators. If the arch moderators feel the need to remove the package because I don't check the comments, so be it - but I already have enough on my plate. 

footnote: this is not a passive aggressive rant, its simply an explanation. sometimes tone is hard to pickup.

gwuen:
Thanks for merging and fixing the maintainer section!

Quick question: is the noreply address actually being monitored? The intent of having a contact email in the PKGBUILD is to give users a direct line to the maintainer, not just to satify a formality. It would be worth using one that actually reaches you.

---

I get that maintaining both an upstream project and its AUR package is a lot of work. Still, I think the AUR comment section is worth an occasional check. It's the only channel available to users who don't have a GitHub account, and the submission guidelines do list it as part of the maintainer's responsibilities.

If the time investment becomes too much, having a community member co-maintain or take over the AUR package might be the better option. That's how the vast majority of AUR packages work anyway, and it would free you up to focus on the upstream project.

ejalxndr:
Yes, [noreply@steambrew.app](mailto:noreply@steambrew.app) is a monitored email line, it is linked on our GitHub [home page](https://github.com/SteamClientHomebrew). Bugs are submitted very infrequently, but when they are we see them.

> If the time investment becomes too much, having a community member co-maintain or take over the AUR package might be the better option. That's how the vast majority of AUR packages work anyway, and it would free you up to focus on the upstream project.

Easier said than done. The PKGBUILD is deeply rooted within the Millennium ecosystem, far more than you'd expect; The simplistic look of the source tree is a facade, it's my attempt to make it look more approachable. Maintenance would be a lot more than just updating the build script; patches, features, etc. may need to be incorporated that require knowledge of the entire code base, and all of its SDK. In 3+ years of Millennium's development, I've only come across one other person capable of maintaining the SDKs complexity, and its (one of) the original author(s) @AAGaming00 ([SteamDeckHomebrew](https://github.com/SteamDeckHomebrew)), who frankly understands it better than I do.  As for the core of Millennium, I'm the only one who knows it like the back of my hand. AI maintenance isn't an option either, it just doesn't have enough context to understand everything there is to know about Millennium. All AI contributions have always been slop and unusable. 

Millennium isn't just a generic text editor, file browser, etc. where the development pipeline is already invented and understood beforehand. I'm not inherently saying its hard to learn, just no one cares enough to.

gwuen:
Thanks for the clarification on the email address!

> Maintenance would be a lot more than just updating the build script; patches, features, etc. may need to be incorporated that require knowledge of the entire code base, and all of its SDK.

I see your point, but I think there's a distinction worth making. I'm not suggesting that someone outside the project takes over the upstream build infrastructure. The cmake presets and everything tied to the codebase obviously needs to stay with you.

What I was referring to is the AUR package specifically: bumping versions, monitoring comments, and filing issues upstream when something is broken. That doesn't require SDK knowledge, just someone who is responsive to user feedback.

The two roles don't have to be the same person, and for most AUR packages of projects this size, they aren't.

PR #821, opened 2026-06-26
Starlight compiler toolkit support

### LINKED ISSUE #683: [Feature] Ability to export logs of all chromium pages for easier debugging
### Before Requesting

When something errors in a plugin's frontend (like the `SharedJsContext` or other pages), it's really hard to get those logs from users experiencing the issue. Right now they have to:
- Launch Steam in dev mode
- Navigate to localhost:8080
- Find the correct page
- Open the console tab (which is sometimes hidden)
- Right-click and export

This is complex, doesn't give you all the pages and many users struggle with these steps.

It would be really nice if in the logs tab there could be an extra button that when clicked downloads a zip containing text files or json of the logs from all frames like the `SharedJsContext` and any open BrowserViews. This would make it way easier to debug errors that are only happening to some users and are hard to replicate.

The logs could be easily gathered with the Chrome Devtools Protocol by attaching to each page, listening to the `Runtime.consoleAPICalled` event and calling `Runtime.enable` which will then fire the event for all previous logs.

This would be an huge quality of life improvement when debugging user errors.

I currently can't think of the right way to fully implement this, that's why I am leaving it here as a feature request.

### Anything else?

_No response_

shdwmtr:
# Starlight compiler toolkit support

This PR adds support for the starlight compiler plugin format (`.star` binaries) alongside the existing loose-file (v1) plugin format. v1 plugins are fully backwards-compatible and continue to work without changes.

```toml
[plugin]
id = "com.example.boilerplate"
name = "Boilerplate Plugin"
version = "1.0.0"
author = "Millennium"
description = "A template plugin for Millennium."

[assets]
# reference resources to be included inside the plugin at compile time
resources=["assets/image.png"]

# Debating whether to introduce this, or just use static checks:
# A simple script that returns a tuple [boolean: ok, string: error].
# ok = plugin is usable on the current platform/environment
# error = the error message if !ok
# ex: jit.os == "Windows"
# [preload]
# entry = "preload.lua"

[backend]
entry = "backend/main.lua"
sources = ["backend/**/*.lua"]

[frontend]
entry = "frontend/index.tsx"

[frontend.globals]
# ex: 
# "@steambrew/client" = "window.MILLENNIUM_API"

[webkit]
entry = "webview/preload.ts"

[webkit.globals]
# same as frontend.globals

[compiler]
# output path
# auto = automatically detect Millennium installation and install plugin
# any other path = used as the output path for the build
output_path = "auto"

[dev]
# plugin_name = "com.example.boilerplate"     # defaults to [plugin].id
# socket = "/tmp/millennium-mep.sock"         # defaults to platform MEP path

# Restart the plugin backend on any source change (default: true).
auto_restart = true

# Reload Steam's UI frontend after a restart, depending on which files changed:
#   "never"    = never reload the UI (default)
#   "backend"  = reload when .lua files changed
#   "frontend" = reload when .ts/.tsx/.js/.jsx/.css files changed
#   "both"     = reload on any source change
reload_steamui_when = "both"

[inspect]
depth = 4
colors = true
show_hidden = true

[logger]
# padding = -15                   # negative = right-align (pad left), positive = left-align (pad right), 0 = none
# backend_prefix  = "BCK {file}"  # {file} expands to "filename:L{line}" when available
# frontend_prefix = "FRT {file}"
# webview_prefix  = "WEB {file}"
# backend_col  = [160, 90, 180]
# frontend_col = [90, 148, 200]
# webview_col  = [65, 165, 160]

```

## High Level
```bash
# v1 plugin, loose files 
~/.local/share/millennium/plugins/Extendium/**/*

# v2 plugin, compiled .star binary
~/.local/share/millennium/plugins/com.BossSloth.Extendium.star
```

### The disconnect
Many new community members find the disconnect between the frontend/webkit and backend confusing. It's not necessarily common sense that the frontend needs to be recompiled, and can be reloaded at runtime with F5, but backends don't need compilation and require a full Steam restart. 

### Where do I develop?
v1 plugins either have to be sym/hardlinked to Millenniums plugin folder, or directly in it. This is shitty. Millennium should only need the final binary, nothing to do with the plugins source code. 

### Hot Reloading (HMR)
Hot reload has always been an untouched mess. It's non-existent. With the introduction of `starlight`, we can easily perform HMR with Millennium over the MEP communication wire.  

### Where tf are the logs, and how do the users find them?
No one had the answer. `starlight` fully addresses this: all logs (backend, frontend, webkit) output flat to `starlight` in watch mode over the MEP using a configurable `util.inspect` polyfill from NodeJS. Without a debugger attached, logs still go to their default location WITH the frontend and webkit logs. 

TODO: 
* Allow `millennium.toml` to configure log format when a debugger is active, vs their default output when a user is running them.

## `.star` Documentation

```
Layout:
  0      u8    id
  1      u8    encode_flags   (COMPRESSED=0x80, OBFUSCATED=0x40)
  2–3    u8[2] pad            (zeroed)
  4–7    u32   meta_flags     (DEFERRED=0x01, REQUIRED=0x02)
  8–15   u64   offset         (bytes from star_start, LE)
  16–23  u64   length         (encoded on-disk byte count, LE)
  24–27  u32   crc32          (of encoded data)
  28–255 u8[]  reserved       (zeroed; future fields go here)
```

A .star archive is a self-contained binary *driver* for a Millennium plugin. While compiling, `starlight` generates an internal shim that only itself understands. Millennium simply asks the plugin to bootstrap itself, the plugin parses its sections, assets, and metadata itself - completely decoupled and independent from Millennium. It packages the backend, frontend, webkit, resources, and metadata into a single binary with:

- LZ4 compression with optional obfuscation
- Parity integrity checks: `xor` && rolling chain per block
- Ed25519 sig support (for signing trusted plugins on the PluginDatabase CI)
- Zero-copy asset lookup at runtime without extracting to disk

### Lua API Changes
For the "resources" section in `millennium.toml`

```lua
---Asset sub-module for reading files bundled into the .star at compile time.
---Data is never preloaded; every call opens the .star file and reads only the requested entry.
---@class millennium.assets
millennium.assets = {}

---Read the raw bytes of a bundled asset.
---Decompresses on the fly. Returns nil if path is not found.
---@param path string Asset path as it appears in the .star (e.g. "assets/icon.png")
---@return string|nil data Raw file contents, or nil if not found
function millennium.assets.read(path) end

---Return the uncompressed size of a bundled asset without reading its data.
---@param path string Asset path
---@return integer|nil bytes Uncompressed size in bytes, or nil if not found
function millennium.assets.size(path) end

---Return the base name (last path component) of a path.
---@param path string Asset path or directory path
---@return string name The final path component, with no trailing slash
function millennium.assets.name(path) end

---Return whether a path points to a "file" or "directory" inside the asset bundle.
---@param path string Asset path (no trailing slash)
---@return "file"|"directory"|nil kind Type of entry, or nil if not found
function millennium.assets.type(path) end

---List the direct children of a directory inside the asset bundle.
---File entries are returned as plain names; directory entries have a trailing "/".
---@param dir? string Directory path (default: "" for root). Trailing slash is stripped automatically.
---@return string[] children Sorted list of direct children
function millennium.assets.list(dir) end
```


ejalxndr:
Thanks co-pilot.

PR #828, opened 2026-07-01
fix(LuaAPI): use lua_type for type checking to avoid coercion

DeveloperMikey:
Due to [coercion](https://www.lua.org/manual/5.1/manual.html#2.2.1) lua_isstring returns 1 on both strings and numbers, and lua_isnumber on numbers and strings that are convertible to numbers. [This is clearly documented](https://www.lua.org/manual/5.1/manual.html#3.7)

This means passing a number to call_frontend_method as a function argument results in a string being sent to the frontend function instead.

Comparing types directly should be a much better way to do it.

### DIFF
--- a/src/lua_host/api/millennium.cc
+++ b/src/lua_host/api/millennium.cc
@@ -89,15 +89,19 @@ static int RPC_CallFrontendMethod(lua_State* L)
         int len = static_cast<int>(lua_objlen(L, 2));
         for (int i = 1; i <= len; ++i) {
             lua_rawgeti(L, 2, i);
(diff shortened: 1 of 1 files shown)


ejalxndr:
Unfortunately, if this has been an issue; all plugins have already worked around it. If we fix it now, it inadvertantly re-breaks already broken plugins. 

I recommend checking out the latest plugin template, which uses a newer engine. It shouldn't have the same issues (hopefully): 

https://github.com/SteamClientHomebrew/PluginTemplate/tree/starlight-v2-support

Legacy compatibility is more important than correctness here.

PR #664, opened 2026-03-04
Feature/macOS support based on async-ffi branch

iFleey:
CleanShot 2026-03-05 at 01 12 16@2x
CleanShot 2026-03-05 at 01 12 36@2x
CleanShot 2026-03-05 at 01 12 58@2x

This PR adds support for both Apple Silicon and Intel-based Macs.

Core functionality has been verified on both architectures, with no major issues observed.

Tested Devices

- MacBook Air (M4, Apple Silicon)
- MacBook Pro (2016, Intel)

Thanks for taking the time to review it!

PR #841, opened 2026-07-12
feat(macOS): Restore macOS support

iFleey:
Hey again! The macOS support from #664 shipped in v3.0.0, but it quietly rotted on `main`. With no macOS CI to catch it, the async-ffi / sdk-flatten / pipe-CDP refactors reintroduced a handful of Linux-isms and left the CDP transport a `#else` stub — so on `main` today it doesn't build, and even patched locally the menus won't open. This PR gets it green again.

Roughly whats in here:

- Build drift
  a few Linux-only assumptions (endian headers, inotify, `SOCK_CLOEXEC`, a Rust return type) crept back in without `__APPLE__` branches. Restored them.

- CDP transport
  this was the real "menus won't open" cause. After the move to a pipe-based transport, macOS was left a stub with its old WebSocket path deleted. macOS is different: the `steam_osx` wrapper enables CEF remote debugging over TCP, so I implemented the macOS path to talk CDP over a libcurl WebSocket (re-resolving the debugger URL each attempt so a steamwebhelper restart doesn't strand it). First clean run: menus open, `Connected to SharedJSContext in 25 ms`.

- Shutdown crash
  remember the `mutex lock failed: Invalid argument` on shutdown from #664? macOS only now runs to a clean `exit()`, which exposed one root cause: a static destruction-order issue where teardown locked singletons `__cxa_finalize` had already destroyed. Leaking them keeps the sync primitives alive past teardown. Shared code, so it helps every platform.

- Scripts
  `scripts/macos/*` paths drifted after the layout changes; fixed those too.

I also built a macOS CI job + bundle-based install on a branch stacked on this one — [`macos-dist`](https://github.com/iFleey/Millennium/tree/macos-dist). Since it's based on this PR I'm holding off on a second PR, but it's fully green if you want to peek: [`Github Actions`](https://github.com/iFleey/Millennium/actions/runs/29184031629), build-macos passes; the only red job is Windows code-signing, which needs a repo secret my fork doesn't have.

### Tested Devices
- MacBook Air (M4, Apple Silicon)

I only have Apple Silicon on hand this round,so it's **arm64-only** and I couldn't test x86_64. Steam runs arm64 natively onApple Silicon so this is the right slice here.

Thanks for taking a look!

CleanShot 2026-07-12 at 15 54 56

CleanShot 2026-07-12 at 15 55 07

CleanShot 2026-07-12 at 15 55 28

CleanShot 2026-07-12 at 15 55 33

CleanShot 2026-07-12 at 15 55 45

> BTW, actually the Extendium plugin does not support the Mac platform.
```
[45:48.483] EXTENDIUM ========================================
[45:48.483] EXTENDIUM [install] Starting extension installation...
[45:48.483] EXTENDIUM ========================================
[45:48.483] EXTENDIUM [install] Extension directory: ~/Library/Application Support/Millennium/plugins/extendium/fake-header-extension
[45:48.483] EXTENDIUM [install] Reading extension-keys.json: ~/Library/Application Support/Millennium/plugins/extendium/fake-header-extension/extension-keys.json
[45:48.483] EXTENDIUM [install] Extension ID: ilceigfjdfkplpmafppofppbifclfdop
[45:48.484] EXTENDIUM [install] Reading manifest.json: ~/Library/Application Support/Millennium/plugins/extendium/fake-header-extension/manifest.json
[45:48.484] EXTENDIUM [install] Extension: Extendium - Fake Header Extension v1.0.0
[45:48.484] EXTENDIUM [install] Steam path: ~/Library/Application Support/Steam/Steam.AppBundle/Steam/Contents/MacOS
[45:48.484] EXTENDIUM [install] Platform: Linux
[45:48.484] EXTENDIUM [install] Steam config directory: ~/.steam/steam/config/htmlcache/Default
[45:48.484] EXTENDIUM [install] Steam config directory does not exist: ~/.steam/steam/config/htmlcache/Default
```


src/engine/cdp_connector.cc
```diff
@@ -362,6 +362,264 @@ void socket_utils::connect_socket(std::shared_ptr<socket_utils::socket_t> socket
...
+#include <mutex>
+#include <memory>
+#include <curl/curl.h>
+#include <curl/websockets.h>
+#include <nlohmann/json.hpp>
+#include "millennium/millennium_lifecycle.h"
+#include "millennium/http.h"
+
+/*
+ * macOS uses a TCP remote-debugging PORT rather than the fd pipe transport that
```
ejalxndr: > macOS uses a TCP remote-debugging PORT rather than the fd pipe transport that Windows/Linux use (--remote-debugging-pipe)

Is there a specific reason we do this?

ejalxndr:
Omfl, I left a little review hours/days after this PR was opened, and I was wondering why I never got a response

image

I assumed "pending" meant I was awaiting you, I was unaware this was even possible 🤦

image

iFleey:
> Omfl, I left a little review hours/days after this PR was opened, and I was wondering why I never got a response
> 
> image
> I assumed "pending" meant I was awaiting you, I was unaware this was even possible 🤦
> 
> image

Haha no worries, thanks for coming back to this!

src/engine/cdp_connector.cc
iFleey: I kept TCP because the macOS wrapper already had it working, and I was focused on getting macOS support back up. I went back and tested pipes in a full Steam session. Connecting to SharedJSContext worked, and after killing the Helper, Steam restarted it and I could connect again through a new pipe.

So yeah, there's no macOS limitation forcing us to use TCP here. My comment made it sound that way. The pipe test used a separate probe, so I still need to wire it into Millennium and test the plugins, but I think switching makes sense.

iFleey:
Switched this over to pipes. I tested with Millennium loaded this time, then killed the Helper and let Steam restart it. SharedJSContext connected again and the SDK initialized in the new Helper. HLTB's frontend loaded in my normal session too.

Still seeing a crash when quitting Steam with HLTB making HTTP requests, though. The crashing thread is in OpenSSL while the main thread is waiting for Millennium to stop. Looks like a cleanup-order issue, but I haven't confirmed that yet.

PR #842, opened 2026-07-15
feat: add SteamRT3 client incompatibility warning check

### LINKED ISSUE #840: [Feature] Support for RT3 / 64-bit Steam client
### Before Requesting

**[Feature]: Detect an unsupported RT3 / 64-bit Steam client at install time and warn**

### Problem

When Steam is running the SteamRT3 64-bit client, Millennium installs cleanly, reports success, and then does nothing. There is no error, no log, and no indication to the user that they are in an unsupported configuration.

The reason there is no signal is structural. The hook works by symlinking `ubuntu12_32/libXtst.so.6` to the bootstrap library so it gets preloaded into the 32-bit client process. Under RT3, that process is never launched, so the bootstrap never runs. Millennium is therefore incapable of reporting its own failure. Whatever warning exists has to come from the installer, because nothing downstream of it will ever execute.

Confirmed with the team on Discord that RT3 is not supported and that stable is the focus, which is reasonable. This request is only about making the unsupported state visible instead of silent.

### What the user experiences today

1. `paru -S millennium` (or the install script) completes successfully.
2. Steam launches normally. No Millennium entry anywhere in the UI.
3. No Millennium logs are produced, because nothing ran.
4. The user has no way to distinguish "broken install" from "unsupported client" without reading `install.sh` and comparing the ELF class of the two client binaries.

Step 4 is what I ended up doing:

```
$ file ~/.local/share/Steam/steamrt64/steam
  ELF 64-bit LSB pie executable, x86-64
$ file ~/.local/share/Steam/ubuntu12_32/steam
  ELF 32-bit LSB pie executable, Intel i386
$ pgrep -a steam
  3440 /home/gjp/.local/share/Steam/steamrt64/steam -srt-logger-opened
$ cat ~/.local/share/Steam/package/beta
  publicbeta
```

Both client binaries exist on disk. Only the 64-bit one is executed. That is the whole story, but it takes a while to find.

### Proposed behavior

In `post_install()` (and ideally in the AUR package's install hook), detect an RT3 client and print a warning rather than exiting silently on success. Something like:

```
warn "Detected a SteamRT3 (64-bit) client. Millennium does not support RT3 yet."
warn "The preload hook targets the 32-bit client, which RT3 does not launch."
warn "To use Millennium, disable 'RT3 experimental' in Steam's beta settings."
```

Install can still proceed. The goal is a signal, not a hard block.

### Open question on detection

I do not know where Steam persists the RT3 experimental opt-in, so I cannot propose a reliable check without guessing. `package/beta` containing `publicbeta` is necessary but not sufficient, since RT3 is a separate toggle underneath the beta channel. A few things that might work, and I would rather you pick than have me invent one:

- Whatever config key backs the RT3 experimental toggle, if it is written somewhere readable.
- Presence of a client binary at `steamrt64/steam`. I can confirm this exists on my RT3 install, but I have not verified whether it is absent on a stable install, so this may produce false positives. Someone with a stable client would need to check.

Note also that the Steam root varies (`~/.local/share/Steam`, `~/.steam/steam`, `~/.steam/debian-installation`), so detection should resolve the path the same way the rest of the installer does.

### Environment

- CachyOS (Arch-based)
- Steam 1.0.0.86-2, `publicbeta` channel with RT3 experimental enabled
- Installed via AUR (`paru -S millennium`)

Happy to test a detection patch on this machine, and happy to check behavior against a stable client if that helps pin the check down.

### Anything else?

_No response_

hotline1337:
## Detect unsupported SteamRT3 client and warn at install time

Closes #840 

### What this does

Adds a `check_steamrt3_incompatibility` function that runs after installing to warn the user when Steam is using the SteamRT3 (64-bit) experimental client, since the current preload hook targets the 32-bit client process (`ubuntu12_32/libXtst.so.6`) which RT3 never launches.

As described in the linked issue, without this check the install completes silently and the user has no way to tell "broken install" from "unsupported client".

### Detection logic

Detection is done via three on-disk markers that together indicate an active RT3 session:

- `$STEAM/package/beta` - beta channel is active
- `$STEAM/.steam-enable-steamrt64-client` - RT3 experimental toggle is on
- `$STEAM/steamrt64/steam` - the 64-bit client binary is present

All three must exist for the warning to fire, which avoids false positives on a stable/32-bit install.

### Where it's wired in

- **`.install`** (AUR/pacman hook) - `post_install()` calls the check.
- **`install.sh`** (Other distributions) - same check ported to the shell installer.
- **Nix module** - added to `extraProfile` alongside the existing symlink logic, so it runs in the same shell context where `$HOME` is already correctly resolved for the logged-in user.

Install still proceeds either way - this is a warning, not a hard block, per the discussion in the issue.

### Sample output

```
  WARNING: SteamRT3 client detected

  Millennium does not support the experimental/beta SteamRT3 client.
  For Millennium to work correctly, please either:
  - disable the "Use experimental SteamRT3 Steam Client" option
  - opt out of the Steam Beta entirely
```

### Testing

Tested on Arch with Steam `publicbeta` + RT3 experimental enabled (the environment from the original issue) - warning fires correctly post-install.

### DIFF
--- a/packages/arch/millennium.install
+++ b/packages/arch/millennium.install
@@ -16,6 +16,28 @@ remove_deprecated_patches() {
     fi
 }
 
+# Warn the user if the experimental 64-bit client is enabled
+check_steamrt3_incompatibility() {
+    local steam="/home/$(logname)/.local/share/Steam"
+    local -a required_markers=(
+        "$steam/package/beta"
+        "$steam/.steam-enable-steamrt64-client"
+        "$steam/steamrt64/steam"
+    )
+
+    for marker in "${required_markers[@]}"; do
+        [[ -f "$marker" ]] || return 0
+    done
+
+    echo -e "\e[1;33m"
+    echo -e "\n  WARNING: SteamRT3 client detected\n"
+    echo -e "  Millennium does not support the experimental SteamRT3 client."
+    echo -e "  For Millennium to work correctly, please either:"
+    echo -e "  - disable the \"Use experimental SteamRT3 Steam Client\" option"
+    echo -e "  - opt out of the Steam Beta entirely\n"
+    echo -e "\e[0m"
+}
+
 for_each_steam_user() {
     local callback="$1"
     getent passwd | while IFS=: read -r _ _ uid _ _ home _; do
--- a/packages/nix/steam.nix
+++ b/packages/nix/steam.nix
@@ -24,7 +24,7 @@ let
 
   steamrt3Check = ''
     check_steamrt3_incompatibility() {
-        local steam="/home/$(logname)/.local/share/Steam"
+        local steam="$HOME/.local/share/Steam"
         local -a required_markers=(
             "$steam/package/beta"
             "$steam/.steam-enable-steamrt64-client"
--- a/packages/nix/steam.nix
+++ b/packages/nix/steam.nix
@@ -22,9 +22,36 @@ let
     MILLENNIUM_RUNTIME_PATH = "${millennium}/lib/libmillennium_x86.so";
   };
 
+  steamrt3Check = ''
+    check_steamrt3_incompatibility() {
+        local steam="/home/$(logname)/.local/share/Steam"
+        local -a required_markers=(
+            "$steam/package/beta"
+            "$steam/.steam-enable-steamrt64-client"
+            "$steam/steamrt64/steam"
+        )
+
+        for marker in "''${required_markers[@]}"; do
+            [[ -f "$marker" ]] || return 0
+        done
+
+        echo -e "\e[1;33m"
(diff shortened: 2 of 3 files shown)


ejalxndr:
I think we should just leave the parent issue open you've attached to this. Hard-coding values into build scripts hurts maintainability, especially in externally version-controlled repos - it could block our CI pipeline, requiring manual intervention. Its a headache to roll back when we need to.

If anything, we should just add support instead of blocking it. I just don't have the time to tackle it yet. You seem more than capable of implementing something like this though; given you're interested. Otherwise, we will just let this rot until Steam releases RT3 officially. 

Thank you for your interest in contributing!

PR #847, opened 2026-07-23
fix: Use data-condition attr instead of classes for theme conditions

### LINKED ISSUE #843: [Feature] Normalize class names for options.
### Before Requesting

As of 3.4.0-b9, I noticed that class names are generated for your theme options. However, they are the exact string literal of the option name, which results in multiple class names from spaces. This does not produce a very useful result.

Image

Apostrophes are preserved, which I guess must be valid for class names. I don't know what happens if you have commas or periods etc.

I'd suggest that these class names be normalized into a camel case without the spaces and punctuation. So like `My FREAKin' cool OPTION!` becomes `MyFreakinCoolOption`.

### Anything else?

_No response_

ricewind012:
maybe fixes #843


yurisuika:
A much smarter approach than my idea.

shdwmtr (approved):
Code lgtm

PR #851, opened 2026-07-25
fix: infinite loading after 2026-07-24 steam beta

ricewind012:


shdwmtr (approved):
Code lgtm, I assume you've tested on both stable and beta?

ricewind012:
ok now it works on both

ejalxndr:
Perfect, thanks for the PR. 

Unrelated, I suggest you register/create a PGP key to sign your commits since you're apart of the official org. Just an extra measure to show the user the commit was actually made by you and not forged. It's fairly simple, and puts "Verified" on commits (see above). I never noticed you didn't have it configured.

PR #852, opened 2026-07-26
docs(readme): add auto-generated translation coverage chart

WkdXeqtr:
Small quality-of-life thing. Adds a translation coverage chart to the README (under Language Support) showing how complete each locale is against english.json, plus a workflow that regenerates it whenever the locales change. No third-party service, it just counts keys.

Handy for contributors to see at a glance which languages need work, and it is the same idea I would reuse for plugin localization down the line.


WkdXeqtr:
Sorry, pushed the wrong version first. Swapped it for an SVG chart - I think it looks better. It's served per theme with `<picture>`, so it renders correctly in both light and dark.

shdwmtr (approved):
Looks great!

Issue #102, opened 2024-09-12
Ability for themes to rely on certain plugins.


Plaer1:
### Before Requesting

- [X] I found no existing issue matching my feature request
- [X] This request is related to Millennium, and not a request for a specific theme/addon

### Describe the feature you'd like!

I currently run the ATOMS skin and I'm going to put the ITAD integration into it's own plugin for everyone to use soon ʸᵃʸ. 
I would like to be able to have this plugin load **_when users install my theme_**, that way when you install my theme it still "feels" like the ITAD integration is a feature of the theme.

I know better discord has a dependency system for the plugins, I think having something similar will be good in the long run .I'm sure people will end up making libraries for their plugins eventually as well, I just think it would be cool to have it where themes could be dependent on plugins (but not the other way around, I do-not want to encourage a "walled garden" situation with the plugins).
![2024-09-12_1726161283](https://github.com/user-attachments/assets/fda0635d-8123-440d-babf-a414fb8b9465)

PR #855, opened 2026-08-02
feat: add plugin dependencies

### LINKED ISSUE #823: [Feature] Add plugin dependencies (prerequisite for third-party localization)
### Before Requesting

You said a plugin localization system could work but would need proper planning, so here it is. The short version: I want third parties to be able to ship translations for plugins they don't own, and for that a plugin needs to be able to depend on another and load after it. So dependencies come first, localization builds on top. This is the dependency design - I want your take before I write any code, and it would land in both Millennium and the PluginTemplate.

Add an optional `dependencies` to plugin.json - a list of internal plugin names (the same `name` the loader already uses), with an optional semver range since CMP_VERSION already exists:

```jsonc
"dependencies": ["steam-some-plugin", "another-plugin@>=1.2.0"]
```

On install. No auto-installing anything - same stance you took on the sidebar PR. If a plugin has missing dependencies, the installer shows a dialog listing them and lets the user install anyway or cancel. The plugin is expected to degrade gracefully if a dependency is absent.

Resolution. Check what's actually on disk, so a dependency someone installed manually counts. Three states: missing, present but disabled, present and enabled. Each dependency name is resolved to its page via the same plugin API the installer uses, shown as a link; if the lookup fails, show the raw id.

Load order. Backends already all start before any frontend, but the order among them is arbitrary, so I'd topologically sort the spawn order (cycle detection falls back to current order with a warning). Frontends have no ordering at all today since the shims are appended as async module scripts, so I'd sort them in dependency order and make injection ordered. Plus a `Millennium.whenPluginReady(name)` as a safety net.

`openQuickAccess({ plugin })` fits here as a compatibility affordance: a plugin can deep-link into a dependency's settings panel, including from the dependency dialog.

How localization uses this (for context): a plugin ships a `locales/` folder with `english.json` as the source of truth, plus a small helper in the SDK (same pattern as the core locale manager). A translation plugin then depends on the target and contributes its strings to a shared registry in core - it has to be in core because plugins run in isolated worlds and can't patch each other directly. The target's lookup merges its own strings with any contributed overrides. No target installed means the translation plugin is a no-op. I'm not trying to extract strings from old plugins; a plugin is localizable once it adopts the system.

A few open questions:

- Dependencies referencing internal `name` with optional `name@range` - good, or bare names only for v1?
- Warn-and-allow vs hard block on missing deps. I lean warn.
- The ordered frontend loading touches the loader and shim injection - want me to take a pass at it, or own that part yourself like you did with quick access?

### Anything else?

References for context:

- `Millennium.openQuickAccess({ plugin })` from the quick access work in `next`.
- Version comparison already wired through the `CMP_VERSION` IPC in the plugin loader.
- The core locale manager (`src/typescript/frontend/utils/localization-manager.ts`) as the pattern for the per-plugin helper.

WkdXeqtr:
Closes #823 and #102.

Adds an optional "dependencies" field to plugin.json (and skin.json for themes):

```
"dependencies": ["some-plugin", "other-plugin@>=1.2.0"]
```

- plugins load dependency-first. Backend spawn is sequential and waits for each on_load, so the order is a real guarantee. Cycles warn and fall back to scan order
- installing a plugin or theme with unmet dependencies shows a warning listing each one with its state. The user can always continue, nothing is ever installed automatically
- plugin rows in settings get a small chip per dependency with a check or cross

Installing a plugin with unmet dependencies

Same dialog for a theme

Dependency chips in the plugins list

Without the field nothing changes (covered by a test). Version ranges support >=, <=, >, <, =.

This also makes sharing between plugins practical: a plugin can expose objects or CSS variables (shared panel colors and such) and a dependent can actually rely on them, since its dependency is guaranteed to be present and loaded first. Same groundwork the locale system from #823 would build on.

Would like your take on one thing: this is deliberately soft - a missing dependency never blocks installing or enabling, authors are expected to degrade gracefully. If you'd rather also have a hard variant (can't enable without it, npm dependencies vs peerDependencies style), that can be added on top - happy to rework in either direction.

Not here: ordered frontend script execution and a ready signal for plugin frontends, can follow up. Note the install dialogs read pluginJson / skin_data from the site API, which forwards the manifests as-is today.


ejalxndr:
Big fan of the incremental implementation of this FR. It makes it very easy for me to review. 

A hard dependency variant sounds viable. 

## Major 

1. The dependency-warning log only ever fires once, total, for the entire Steam session (not once per problem as it should).

## Minor 

1. we can dedup the O(n) name lookup in the warning loop instead of reusing `index_by_name` from `plugin_deps.cc`
2. `DependencyVersionLabel` collapses > and >= into identical UI text ("min version X"), which loses distinction that semver::satisfies actually enforces.
3. plugin_deps.h uses size_t without `#include <cstddef>`. works rn via transitive includes, but not guaranteed.
4. Dep chips show config-enabled, not actual running state (`PluginComponent.tsx:597`). A crashed backend still counts as "enabled" in config, so a dependency chip can show "✓" while the actual guarantee the PR advertises (loaded first) doesn't hold.

WkdXeqtr:
Thanks, all five addressed:

1. warnings are deduplicated per problem (plugin + dependency + reason) rather than latched for the whole session, so each distinct issue is reported once no matter how often the list gets sorted
2. sort_by_dependencies builds the name index once through plugin_deps::index_by_name and shares it with the warning pass
3. >= and > (and <= / <) render differently now - "min version 1.2.0" vs "newer than 1.2.0"
4. chips treat a crashed dependency as unmet, since the config still says enabled but nothing is actually loaded
5. added the include

Also took a shot at the hard variant, in its own commit so it can be dropped independently. "requires" sits next to "dependencies", npm style:

    "requires": ["some-plugin"],
    "dependencies": ["nice-to-have"]

Both order the load, but a required one is enforced: a plugin can't be enabled until its requirements are there, and a theme that lists requirements can't be applied. The toggle stays off with a tooltip naming what's missing, and the native side refuses the same requests so a hand-edited config can't sneak past. Turning off a plugin others require asks first and turns those off too, dropping back to the default theme if the active one required it - otherwise the guarantee would only hold until someone flipped the dependency off.

Installing stays permissive either way, the dialog just labels those entries "required", since installing something you can't enable yet is harmless.

Rebased onto current next.

WkdXeqtr:
Marking this ready - the review notes are addressed and the hard variant is in as its own commit. I can split it up if the branch is too much at once.

WkdXeqtr:
Realised the dependency work only covered v1 plugins, so this adds the v2 side: starlight writes "requires" and "dependencies" from [plugin] into the .star metadata, and Millennium reads both back out. Before this it only read "dependencies", so a required dependency was silently dropped for star plugins.

The [plugin] config fields also show up in #887 for the localization side. Whichever lands first, I'll rebase the other.

ejalxndr:
Honestly given that, maybe we should drop support entirely from V1 plugins, and only have it as a V2 feature? Would be much less to maintain. I'm open to discussion!

WkdXeqtr:
Works for me - new plugins that want to declare dependencies can be expected to use the new format anyway, so I'll strip the schema change and the plugin.json reads.

Two things I'd keep: v1 plugins staying valid as targets, since the resolver matches by name whatever the format and an old plugin that never migrates can still be depended on - that lives in the shared core and costs nothing. And themes, since skins have no v2 format - that side reads skin.json either way.

PR #859, opened 2026-08-07
fix(locales): localize theme installer status messages

WkdXeqtr:
Was messing around with themes and noticed the installer progress is still hardcoded English - "Downloading theme...", "Setting up theme...", "Done!" and so on, all the way through install and update. Same for the update toasts ("Updates Available" / "We've found N updates").

Did the same thing we did for the plugin installer earlier: the C++ side dispatches `##strSomeKey` and the frontend resolves it, which it already does for theme progress too.

- theme_mgr.cc - all 14 progress strings for install and update
- plugin_mgr.cc - two leftovers in the update path that got missed last time
- the library update toast now reuses `themeAndPluginUpdateNotification` + `updateSingular`/`updatePlural`, which were already in english.json but unused by anything. Slightly different wording than the hardcoded string, but it also stops saying "1 updates"


shdwmtr (approved):
Looks great!

PR #860, opened 2026-08-10
fix(frontend): support Steam beta root menu

mdamer12:
## Summary
- discover the Steam root menu React tree from either the legacy `#root` mount or the current beta `#popup_target` mount
- retry briefly while the popup React tree is mounting
- retain the normal React patch and add a cached-render fallback for Steam beta popups
- reuse the same localized menu item definitions and remove fallback nodes if a later React render takes over

## Problem
Steam public beta build `1786141909` pre-renders and reuses the root-menu popup below `#popup_target`. Millennium's `getReactRoot(document.getElementById('root'))` lookup therefore returns no tree, leaving the backend and plugins loaded but the Millennium menu unavailable.

## Testing
- `bun run typecheck` in `src/typescript/frontend`
- complete Windows release build with CMake/MSVC
- live-tested on Steam public beta build `1786141909`
- verified one Millennium menu pair, Settings navigation, menu persistence after reopening, and successful backend/frontend plugin loading

### DIFF
--- a/src/typescript/frontend/patcher/index.ts
+++ b/src/typescript/frontend/patcher/index.ts
@@ -172,7 +172,7 @@ export async function onWindowCreatedCallback(windowContext: any): Promise<void>
 
 	/** Patch the steam root menu to add the Millennium root menu */
 	if (windowTitle === 'Steam Root Menu') {
-		PatchRootMenu();
+		PatchRootMenu(windowContext.window.document);
 	}
 
 	const prevMainWindow = pluginSelf.mainWindow;
--- a/src/typescript/frontend/settings/index.tsx
+++ b/src/typescript/frontend/settings/index.tsx
@@ -210,8 +210,14 @@ export function MillenniumSettings() {
 	);
 }
 
-function RenderSettingsModal(_: any, retObj: any) {
-	const items = [
+export type SettingsMenuItem = 'separator' | {
+	name: string;
+	onClick: () => void;
+	visible: boolean;
+};
+
+export function GetSettingsMenuItems(): SettingsMenuItem[] {
+	return [
 		{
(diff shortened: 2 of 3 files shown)


ejalxndr:
What exactly is this fixing? I can't reproduce the issue on beta - Millennium shows in the context menu. Could you include images/videos? I tested the latest beta `1786661881`

mdamer12:
Thanks for testing. This PR originally addressed the root-menu mount change I reproduced on beta build 1786141909, where the menu was under #popup_target and Millennium’s #root lookup failed. If the menu is consistently available again on 1786661881, that frontend issue may no longer reproduce on the latest beta.
I also found a separate issue with official Millennium v3.4.0 on 1786661881, but it is unrelated to this PR. Steam crashes during startup with 0xC0000005 at tier0_s64!CreateSimpleProcess+0x4e, called from millennium.dll. The v3.4.0 hook forwards three arguments, while the current function receives a fourth working-directory pointer. In my dump, r9 contained 0x3, which Steam attempted to dereference. I’ll keep that as a separate fix/report rather than mixing it into this frontend PR.

Steam beta: 1786661881
Millennium: official v3.4.0

ExceptionCode: c0000005 (Access violation)
Fault: tier0_s64!CreateSimpleProcess+0x4e
Attempted read address: 0x0000000000000003
r9: 0x3

Call stack:
tier0_s64!CreateSimpleProcess+0x4e
millennium+0xeae5b
SteamUI

ejalxndr:
Indeed, I fixed that in [v3.4.1-beta.1](https://github.com/SteamClientHomebrew/Millennium/releases/tag/v3.4.1-beta.1) yesterday! Not sure why they decided to change it after 9 years!

PR #865, opened 2026-08-11
Update german.json

kajoken:
Changed from formal to informal German since it felt kinda wrong (personal preference) and the first translation switched from formal to informal several times as well. So it's also for consistency. Changed the flow of some existing phrases.

**Explanation of some of my choices.**

* Plugins = We use this in German as well, Alternative words could be "Software-Erweiterung" or "Zusatzmodul", which both sound kinda meh if you ask me.
* Themes = Designs, even though that's also English. Using "Theme" in German wouldn't be a dealbreaker though. This comes down to personal preference.
* Quick CSS = There's no real German counterpart for this. Lets keep it like that. If you really want to translate that, you'd get "schnelle kaskadierte Stylesheets" and ... no, just no.
* Updates = The German word for Update would be "Aktualisierung", but AFAIK most tech savvy Germans would also use "Update".
* Client Theme = I used "Programmdesign" here. You could also use "App Design" as well.
* Updating... = Even though we also use Update, we don't use the verb. Instead it would be "aktualisieren" coming from "Aktualisierung" or like here: "Wird aktualisiert". We also could use the same style like the original which would give us "Aktualisieren...".
* Stable Channel = There's no direct translation here. I choose to take "Hauptkanal", which would translate to "main channel". I think that's close enough. "Stabiler Kanal" would be the word for word translation, which sounds very weird for Germans (probably).
* Show Notification Toasts = I decided to go with "Pop-up-Benachrichtigungen" instead of "Push-Benachrichtigungen" like the other translation. Mainly because a Push Notification is different from a Toast. We don't have a true translation for "Toast" in this context though.
* Whoops! = I went for "Hoppla!", "Ups!" is also very valid, again personal preference.
* Your issue is not valid = I decided to go with "Dein Problem ist nicht stichhaltig" instead of "Dein Problem ist ungültig". The latter would be the word for word translation but in context sounds harsher than it probably should be.
* View Download Information = "Informationen zum Download anzeigen" another one of those English words that have been assimilated into German. The word for word translation would be more like "Informationen zum Herunterladen anzeigen". Not really a fan of that.
* valid build/invalid build = We use "Build" in this context as well. One could also translate it to "gültige/ungültige Version", but Version and Build may not be 100% the same thing here.
* Never Mind = I honestly may be wrong here. Since I did not 100% understood the context. I went with "Macht nichts" but it could also be "Vergiss es". Those two have a very different tone though. The first could also be translated to "no problem" in a more friendly way where the second is a harsher "forget it".
* Proxy = Is *drumroll* ... Proxy. Yep, another one of those. I think we don't have anything different for that at all.
* Loader build date = To be honest I really didn't know how to translate that. "Loader" and "Build" can both be used in German as well in the context of IT. So I went for "Loader Build Datum".

I'm of course open to any suggestions. Feel free to point out errors, be it spelling or structure.


yurisuika:
Some key/value pairs taken from the actual Steam localization files for best consistency...

`Updating`
- Updating...
- Wird aktualisiert …

`Audio_Reset_Config_Confirm_Description`
- This will erase all audio settings and custom configuration, including those installed by other software and plugins. The system will restart after this operation.
- Dadurch werden alle Audioeinstellungen und benutzerdefinierten Konfigurationen gelöscht, einschließlich derer, die von anderer Software und Plug-ins installiert wurden. Das System wird nach diesem Vorgang neu gestartet.

So, "Plug-ins" may be a good choice for to use.

`NotificationsSettings_ShowNotifications`
- Show Notification Toasts...
- Pop-up-Benachrichtigungen anzeigen …

That's just a few examples. Hope this can be some good inspiration.

kajoken:
Hi there. Thanks for the input. Those are some very valid choices. To be honest I was thinking about using "Plug-Ins" first as well but decided to go just with "Plugins" because it can be written both ways in English and German. But you're right, for consistency reason "Plug-Ins" could be better.

Damn, I just looked at the localization file and Steam is using formal German. I guess I need to redo some more parts.

yurisuika:
Also, CSS is not just an abbreviation, it is a file format, so that may probably be best left as "CSS".

kajoken:
> Also, CSS is not just an abbreviation, it is a file format, so that may probably be best left as "CSS".

Yeah, I wouldn't translate that either and used "CSS" in the translation file.

shdwmtr (approved):

